PR #8, opened 2019-07-24
WIP: Country detection helpers

kaedroho:


kaedroho:
No longer required as we don't support regions now.

PR #19, opened 2019-10-31
Update README.md

sanderdatema:
Since the text reads as one phrase, it's confusing what `to` means here. I suppose it should be from (remove from).


kaedroho:
Thanks for submitting this! But I think the current wording is correct, it's asking the user to add a line into their `MIDDLEWARE` setting. Probably should word this more like "Add the following lines into your `MIDDLEWARE` setting" instead.

PR #23, opened 2019-11-11
Store segment HTML attributes in translation memory

kaedroho:


wagtail_localize/segments/ingest.py
```diff
@@ -24,11 +23,11 @@ def organise_template_segments(segments):
         else:
             segments_by_position[int(segment.path)] = segment
 
-    texts = []
+    fragments = []
```
kaedroho: Need to rethink terminology here

wagtail_localize/segments/__init__.py
```diff
@@ -155,10 +155,13 @@ def get_html_element_attrs(self):
         Would return the following dictionary:
 
             {
-                "a1": {"href": "https://mysite.com"}
+                "a#a1": {"href": "https://mysite.com"}
             }
         """
-        return {e.identifier: e.element[1] for e in self.html_elements or []}
+        return {
+            f"{e.element[0]}#{e.identifier}": e.element[1]
```
kaedroho: Need to add tests for these methods

PR #24, opened 2019-11-12
WIP: Implement fallback routing

kaedroho:


alanmoo:
Do you need help testing this?

kaedroho:
@alanmoo Thanks for the offer! I've tested this now (sorry my initial comment was out of date). Routing is working well, what's left to do is reverse routing so that we can generate URLs to these pages that don't actually exist. This is a much harder problem and it needs to run really fast.

The approach I'm thinking of taking is to create a placeholder page in all language trees that don't have the page translated yet. When a placeholder page is served, the fallback routing is triggered. This should make it easier to generate URLs to untranslated pages, even if their parent is untranslated. It'll also be easier to synchronise page moves across languages since we know the destination should always exist.

kaedroho:
Merged into: https://github.com/wagtail/wagtail-localize/pull/47

PR #32, opened 2019-12-17
Treating snippets, images, etc as separate resources

kaedroho:
This PR refactors parts of the segment extraction/ingestion, translation memory and pontoon integration so that related objects are now treated as separate resources.

Previously, these objects were nested on all the pages they appear on. So for example, say we have a page that has a linked snippet, this would be sent to pontoon as a single resource:

```python
SegmentValue("title", "My Page")
SegmentValue("slug", "my-page")
SegmentValue("author_snippet.job_title", "Editor")
```

There's a couple of problems with this:

 - if there are multiple pages that represent the same snippet, the snippet will be translated multiple times and these translations may conflict
 - We need a stable content path for all content so we can use they can be used as the `msgctxt` to allow different translations for different locations (required for https://github.com/mozilla/donate-wagtail/issues/563). Content within snippets currently can't have a stable content path at the moment

This is now broken into two resources

```python
# The page:
SegmentValue("title", "My Page")
SegmentValue("slug", "my-page")
RelatedObjectValue("author_snippet", snippet_object)

# The snippet:
SegmentValue("job_title", "Editor")
```

## Implementation

The reason this was implemented that way is that segments need to be hooked on to an immutable copy of the content so the content can be updated while translations are being made. This only exists for pages at the moment in Wagtail as that's the only model with a revision model.

This PR adds two new models `TranslatableObject` and `TranslatableRevision` which can store an immutable copy of the content for any model. These will also be useful for copying over the mixed synchronised/translated content that exists in streamfields (required to solve https://github.com/mozilla/donate-wagtail/issues/588).


kaedroho:
Superseded by https://github.com/wagtail/wagtail-localize/pull/38 and https://github.com/wagtail/wagtail-localize-pontoon/pull/2

PR #33, opened 2019-12-17
Merge Pontoon integration

kaedroho:


setup.py
```diff
@@ -42,6 +42,12 @@
         'testing': [
             'psycopg2>=2.6',
         ],
+        'pontoon': [
+            'polib==1.1.0',
+            'pygit2==0.28.2',
+            'gitpython==3.0.2',
+            'toml==0.10.0',
```
merwok: Pinned requirements are going to cause issues for apps using this.  Could you use ranges instead?  (like `polib == 1.1.*` or `pygit2 ~= 0.28.2` — both forms allow an app to pin a newer micro version (1.1.1, 0.28.3), the second form specifies a minimum required micro)

kaedroho: Will change this. Thanks for pointing it out!

kaedroho: Fixed (squashed into original commit)

docs/pontoon.md
```diff
@@ -0,0 +1,79 @@
+# wagtail-localize-pontoon
```
kaedroho: This will be rewritten, merging it in so it could be used as a reference for that rewrite.

kaedroho:
Going to leave this separate for now to make it easier to get my other changes merged. I've ported the changes made here back to the original project https://github.com/wagtail/wagtail-localize-pontoon.

PR #34, opened 2019-12-18
Strip out unused code from translation memory module

kaedroho:
I kept this code around for reference, but it'll be out of date after some changes I'm currently making to the translation memory module. Probably best it became git-history now.


kaedroho:
Thanks @nealtodd !

PR #47, opened 2020-01-08
Placeholder pages

kaedroho:
Supersedes: https://github.com/wagtail/wagtail-localize/pull/24

This PR implements a new approach that creates "Placeholder" pages in all locales. These placeholders are synchronised with the best real page for that given locale. For example, `es-MX` will look for `es` first, then `en`.

This means that all locale trees have the exact same structure, so all URLs work and menus can be constructed the same way as on normal Wagtail sites.

Placeholder pages can't be edited, or even have revisions (to simplify the synchronisation). These features get switched back on as soon as a translation is created.

I plan to make it also synchronise when moving pages, but not sure at this time whether we can support synchronising page ordering as well.

TODO:
 - [x] `sync_language_trees` management command
   - [x] Creates missing placeholders
   - [x] Updates placeholder content
   - [ ] Updates placeholder tree position
- [x] Hook for creating placeholders on page creation
- [x] Hooks for publishing/unpublishing placeholders
- [x] Hooks for updating placeholders on page edit/translation
- [ ] Hook for moving placeholders on page move
- [ ] Hook for deleting placeholders


kaedroho:
Superseded By: https://github.com/wagtail/wagtail-localize/pull/191

PR #50, opened 2020-01-20
Add tests for TranslatableMixin

richbrennan:
Adds tests for:

- the TranslatableMixin class

(more to follow)


wagtail_localize/tests/test_mixins.py
```diff
@@ -0,0 +1,108 @@
...
+from wagtail.core.models import Page
+
+from wagtail_localize.models import Language, Region, Locale
+from wagtail_localize.test.models import InheritedTestModel, TestModel
+
+
+def make_test_model(model=None, **kwargs):
+    if model is None:
+        model = TestModel
+    obj = model(**kwargs)
```
kaedroho: Could we use `model.objects.create(**kwargs)` instead?

wagtail_localize/tests/test_mixins.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_has_translation_when_none_exists(self):
+        self.translated_model.delete()
+        self.assertFalse(self.main_model.has_translation(self.another_locale))
+
+    def test_copy_for_translation(self):
+        self.translated_model.delete()
+        copy = self.main_model.copy_for_translation(locale=self.another_locale)
+
+        self.assertEqual("Main Model", copy.title)
```
kaedroho: We should check that the `locale` is correct and `translation_key` of the copy matches the source.

Also should probably check that the copy is not the source instance.

wagtail_localize/tests/test_mixins.py
```diff
@@ -0,0 +1,108 @@
...
+from django.conf import settings
+from django.test import TestCase
+
+from wagtail.core.models import Page
+
+from wagtail_localize.models import Language, Region, Locale
+from wagtail_localize.test.models import InheritedTestModel, TestModel
+
+
+def make_test_model(model=None, **kwargs):
```
kaedroho: We should rename `model` to either `object` or `instance`

kaedroho:
Is this still a WIP?

PR #53, opened 2020-01-28
Remove region and language models

kaedroho:


wagtail_localize/tests/test_locale_model.py
```diff
@@ -0,0 +1,74 @@
...
+            self.assertEqual(Locale.get_active().language_code, "fr")
+
+    def test_get_display_name(self):
+        locale = Locale.objects.get(language_code="en")
+        self.assertEqual(locale.get_display_name(), "English")
+
+    def test_get_display_name_for_unconfigured_langauge(self):
+        # This language is not in LANGUAGES so it should just return the language code
+        locale = Locale.objects.create(language_code="foo")
+        self.assertIsNone(locale.get_display_name())
```
richbrennan: The test comment mentions returning just the language code, but it returns `None`?

kaedroho: Ahh, `__str__` does that, but this should return `None`

Issue #18, opened 2019-10-24
sync_languages output


kaedroho:
This command doesn't currently have any output. Would be nice to see what languages it is activating or deactivating.

kaedroho:
We should probably create `language_activated` and `language_deactivated` signals since this would be quite useful to have anyway. The management command could hook on to these.

kaedroho:
no need for this now that languages are managed in Wagtail core

PR #56, opened 2020-02-11
Feature/18 sync languages output

victoriachan:
For https://github.com/wagtail/wagtail-localize/issues/18

Ideally we want to make use of the Locale `language_activated` and `language_deactivated` signals to display the output for the `sync_languages` command. But not sure how to implement a receiver in the context of a management command. Any suggestions?

### DIFF
--- a/wagtail_localize/management/commands/sync_languages.py
+++ b/wagtail_localize/management/commands/sync_languages.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
+from django.db.models import signals
 
 from wagtail_localize.models import Locale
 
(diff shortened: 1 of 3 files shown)


wagtail_localize/management/commands/sync_languages.py
```diff
@@ -11,12 +12,14 @@ def handle(self, **options):
         language_codes = dict(settings.LANGUAGES).keys()
 
         for language_code in language_codes:
-            Locale.objects.update_or_create(
+            obj, created = Locale.objects.update_or_create(
                 language_code=language_code, defaults={"is_active": True}
             )
+            self.stdout.write("Activated locale: " + obj)
```
kaedroho: I think this should only output if the language was previously inactive

wagtail_localize/signals.py
```diff
@@ -0,0 +1,4 @@
+from django.dispatch import Signal
+
+language_activated = Signal(providing_args=["locale"])
```
kaedroho: Could we rename these to `locale_`?

wagtail_localize/models.py
```diff
@@ -461,3 +459,11 @@ def set_locale_on_new_instance(sender, instance, **kwargs):
         return
 
     instance.locale_id = instance.get_default_locale_id()
+
+
+@receiver(post_save, sender=Locale)
```
kaedroho: This would fire the signals whenever the `Locale` is saved. I guess we probably won't be saving `Locale` for any other reason than to activate/deactivate them, but is there a way for us to detect if `is_active` has changed?

kaedroho: Maybe we could do a check in the `pre_save` signal, but we would still need to fire in `post_save`

Alternatively, we could probably add a separate field that stores the last signalled state? If the field ever defers with `is_active` we know to fire the signal

kaedroho:
Thanks for this! But we've now moved this model into Wagtail core so we no longer have a need for this command

PR #69, opened 2020-05-26
Make TranslationSource.as_instance still work even if object deleted 

kaedroho:
Need to remember why it was done this way in the first place


kaedroho:
No longer required since we changed the way the Pontoon integration works now.

This will still raise an error, but I think it is correct to. We shouldn't create translations after the source has been deleted.

PR #72, opened 2020-06-24
Commands/export import

mardiros:
I'm really happy to find the plugin that manage wagtail i18n using .po[t] file,

Unfornatly the translation request exposed here is not a workflow usefull for my use cases.

At Gandi, we centralize all our translation in a weblate instance which is link to git repositories.

So here, I am adding 2 commands in order that dump all the translations and to reload them back.

I just write it at the moment and I never write a wagtail (or even a django) plugin, so this is experimental

at the moment but I can go on now I have this.

Fill free to give me feedback!


kaedroho:
@mardiros Thanks for this!

I haven't looked at the code yet, but I thought you might be interested in looking at the [wagtail-localize-pontoon](https://github.com/wagtail/wagtail-localize-pontoon/) backend. This is for Mozilla's Pontoon which also uses PO files and Git. This backend handles exporting/importing PO files but it also handles the pushing/pulling a git repository as well.

wagtail_localize/management/commands/import_po.py
```diff
@@ -0,0 +1,121 @@
...
+
+class LocaleLoader:
+    def __init__(self, locale, options):
+        self.locale = locale
+        self.options = options
+        self.items = {}
+        if self.exists:
+            infile = polib.pofile(self.filename, encoding='utf-8')
+            self.items = {
+                entry.msgid: entry.msgstr or entry.msgid for entry in infile
```
mardiros, line 22: `entry.msgstr or entry.msgid` is saving non translated text in the translated string in wagtail db,
otherwise required field constraints failed.

Furthermore, we may use the merge to avoid loosing translations but I am not
sur it is usefull in the workflow.

https://polib.readthedocs.io/en/latest/api.html#polib.POFile.merge

mardiros:
@kaedroho 

Thanks you for the feedback and for sporting `wagtail-localize-pontoon`. 

Right now my little patch is working for my need at the moment, dumping and loading pages, but it may not scale actually,
I don't know at the moment.

wagtail_localize/management/commands/export_po.py
```diff
@@ -0,0 +1,152 @@
+import pathlib
```
kaedroho, line 1: I think these commands belong in the `translation` sub-app. The idea of the sub-app is that you can disable all the translation stuff if you only wanted to use the mixins but none of the translation/PO file stuff.

wagtail_localize/management/commands/export_po.py
```diff
@@ -0,0 +1,152 @@
...
+
+import polib
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.utils import timezone
+from wagtail.core.models import Site
+
+from wagtail_localize.models import Locale, default_locale_id
+from wagtail_localize.workflow.views.translate import MessageExtractor
```
kaedroho, line 10: Just to note that this code is just a stop-gap until I implement something proper at the model level. Happy for it to import these from views for now, but we will probably end up rewriting some of the logic here once the real API for extracting/ingesting strings has been developed

kaedroho:
> Unfornatly the translation request exposed here is not a workflow usefull for my use cases.

Out of interest, what is it missing? Is it the having to use a web interface? or something about the way it worked?

I plan on making it return a zip file soon with an actual folder structure (instead of one giant combined PO file), would that help in your case?

wagtail_localize/management/commands/export_po.py
mardiros: I am sorry but I am really not sure to understand you are talking about.

Are you suggesting to keep those Commands in my sub-app ?

wagtail_localize/management/commands/export_po.py
mardiros: Yes, I don't want to refactor all of that and I agree that importing stuff from the view is not a correct design.

mardiros:
> > Unfornatly the translation request exposed here is not a workflow usefull for my use cases.
> 
> Out of interest, what is it missing? Is it the having to use a web interface? or something about the way it works?
> 
> I plan on making it return a zip file soon with an actual folder structure (instead of one giant combined PO file), would that help in your case?

First, my usecase is a bit specific, I am not going to use wagtail to build a website, but to create Jinja2 templates.
Yes it is a bit meta, I am giving a try to use wagtail as a headless CMS to store notification template such as email.
Using wagtail make it is easy to create another email template by instanciating a new page.

For me the first constraint was to not introduce another web interface for our translators in order to validate
with my collegue that we can still use weblate to translate.

Apart from that, I may have to import translations from the current system to this wagtail instance, so I will need scripts for that.

So, the web interface is not what I am looking for right now, but maybe we can use it in the future, if our product team and
translator team are ok with that afterall. I am just a backend developper :wink:

wagtail_localize/management/commands/import_po.py
```diff
@@ -0,0 +1,135 @@
...
+            translations = LocaleLoader(target_locale, options)
+            message_ingestor = MessageIngestor(
+                source_locale, target_locale, translations.items,
+            )
+
+            for page in get_pages(root_page):
+                print("Loading page {}".format(page))
+                instance = page.get_latest_revision().as_page_object()
+                if instance.is_source_translation:
+                    if instance.locale.id != source_locale.id:
```
mardiros, line 126: I have an issue with that, in fact the pages are localized, and I don't find how it works, is it the `LANGUAGE_CODE` from the settings when it the webpage was created ?

I don't find any way to customize it. I want to keep the `en` as the django lang for my command line and for the admin,
but not for the wesite published (`sub-app` if it is called like this).

wagtail_localize/management/commands/export_po.py
kaedroho: Nope, I'm saying put them in `wagtail_localize/translation/management/commands`

kaedroho:
Thanks for submitting @mardiros! Unfortunatly, I'm going to have to close this for now as `wagtail-localize` has been practically rewritten since it was opened!

Feel free to resubmit it again if you still want this feature

mardiros:
@kaedroho 

No problem for me. Thank your for your work and your time!

PR #102, opened 2020-07-08
Cleanups and minor tweaks

kaedroho:
Just some minor cleanups that don't fit into a larger PR.


wagtail_localize/translation/segments/extract.py
```diff
@@ -93,6 +93,14 @@ def handle_stream_block(self, stream_block):
 def extract_segments(instance):
     segments = []
 
+    # AIf this is a page, add the parent page as a related object
```
KalobTaulien, line 96: `AIf` -> `If`

PR #111, opened 2020-07-14
Convert <br> tags to newlines and back

kaedroho:
Currently, `<br>` tags are stripped out when converting HTML to text. We should instead convert them to newlines so the spacing in the plain text is correct.


kaedroho:
Replacing with something else

kaedroho:
Replaced by https://github.com/wagtail/wagtail-localize/pull/114

PR #112, opened 2020-07-14
Split out HTML handling logic from SegmentValue into a new class

kaedroho:
This PR introduces a new class called `HTMLSnippet`. This is required in order to allow us to have two HTML snippets per `SegmentValue` so we can have both the source and translation available.

``HTMLSnippet`` represents a snippet of HTML that only includes inline tags.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,46 @@
+version: 2
+
+jobs:
+  backend:
+    docker:
+      - image: circleci/python:3.6.4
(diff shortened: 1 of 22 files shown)


kaedroho:
Still not happy with this. Going to try and simplify it a bit so don't review for now

kaedroho:
Partly replaced by https://github.com/wagtail/wagtail-localize/pull/114

Working on the rest in a separate PR

PR #113, opened 2020-07-16
CircleCI and Flake8

kaedroho:
This PR adds CircleCI config. CircleCI will be used for the following:

 - A fast-responding unit tester (TravisCI has a large matrix so takes a while to respond).
 - Flake8
 - Testing against Wagtail master on a nightly basis (so we can pick up incompatibilities as soon as they happen)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,46 @@
+version: 2
+
+jobs:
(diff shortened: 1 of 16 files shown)


kaedroho:
Thanks!

PR #125, opened 2020-07-21
Machine translators

kaedroho:
This PR Implements two new machine translators:

 - `DummyTranslator` which reverses strings
 - `DeepLTranslator` which translates with [DeepL](https://www.deepl.com)

### DIFF
--- a/wagtail_localize/machine_translators/dummy.py
+++ b/wagtail_localize/machine_translators/dummy.py
@@ -1,0 +1,50 @@
+from bs4 import BeautifulSoup, NavigableString, Tag
+
+from wagtail_localize.strings import lstrip_keep, rstrip_keep, StringValue
+
+from .base import BaseMachineTranslator
+
+
+def language_code(code):
+    return code.split("-")[0]
+
+
+def translate_string(string):
+    # Preserve leading/trailing whitespace. This could be a segment of a sentence.
+    string, left_whitespace = lstrip_keep(string)
+    string, right_whitespace = rstrip_keep(string)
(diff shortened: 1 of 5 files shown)


wagtail_localize/machine_translators/dummy.py
```diff
@@ -0,0 +1,53 @@
...
+    # Preserve leading/trailing whitespace. This could be a segment of a sentence.
+    string, left_whitespace = lstrip_keep(string)
+    string, right_whitespace = rstrip_keep(string)
+
+    words = string.split(' ')
+
+    return left_whitespace + ' '.join(reversed(words)) + right_whitespace
+
+
+def translate_html(html):
```
KalobTaulien, line 23: Would this be a good candidate for a decorator?

wagtail_localize/machine_translators/dummy.py
```diff
@@ -0,0 +1,53 @@
+from bs4 import BeautifulSoup, NavigableString, Tag
+from django.utils.translation import gettext_lazy as _
+
+from wagtail_localize.strings import lstrip_keep, rstrip_keep, StringValue
+
+from .base import BaseMachineTranslator
+
+
+def language_code(code):
```
KalobTaulien, line 9: We have two language code functions in this PR. Any chance we could merge them together?

kaedroho: All the backends have their own special cases when it comes to handling language codes so I think we should keep them separate

PR #133, opened 2020-07-29
Submit translation views

kaedroho:
Views for submitting pages and snippets for translation


kaedroho:
@KalobTaulien Tests are failing because this depends on https://github.com/wagtail/wagtail/pull/6267

wagtail_localize/templates/wagtail_localize/admin/submit_translation.html
```diff
@@ -0,0 +1,47 @@
...
+{% block extra_js %}
+    {{ block.super }}
+
+    <script type="text/javascript">
+        document.addEventListener('DOMContentLoaded', function() {
+            var selectAll = document.querySelector('[name="select_all"]');
+            var locales = document.querySelectorAll('[name="locales"]');
+
+            selectAll.addEventListener('change', function() {
+                locales.forEach(function(locale) {
```
KalobTaulien, line 41: We'll run into IE11 issues with `.forEach()` https://caniuse.com/#search=forEach

wagtail_localize/tests/test_translation_model.py
```diff
@@ -218,7 +218,7 @@ def setUp(self):
         self.fr_locale = Locale.objects.create(language_code="fr")
 
         self.page = create_test_page(title="Test page", slug="test-page", test_charfield="This is some test content")
-        self.source = TranslationSource.from_instance(self.page)
+        self.source, created = TranslationSource.get_or_create_from_instance(self.page)
```
KalobTaulien, line 221: Confirming that `created` doesn't need to be `self.created`, correct?

wagtail_localize/views/submit_translations.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def get_title(self):
+        return self.title
+
+    def get_subtitle(self):
+        return str(self.object)
+
+    def get_form(self):
+        if self.request.method == 'POST':
+            return SubmitTranslationForm(self.object, self.request.POST)
```
KalobTaulien, line 106: Could you use `return SubmitTranslationForm(self.object, self.request.POST or None)` instead of if/else?

wagtail_localize/views/submit_translations.py
```diff
@@ -0,0 +1,212 @@
...
+
+        self.object = self.get_object()
+        return super().dispatch(request, *args, **kwargs)
+
+
+class SubmitPageTranslationView(SubmitTranslationView):
+    title = __("Translate page")
+
+    def get_subtitle(self):
+        return self.object.get_admin_display_title()
```
KalobTaulien, line 174: I dont recall but the admin display title falls back to the page title if it's not explicitly set, right?

wagtail_localize/views/submit_translations.py
```diff
@@ -0,0 +1,212 @@
...
+            page_title=self.object.get_admin_display_title(),
+            locales=locales
+        )
+
+
+class SubmitSnippetTranslationView(SubmitTranslationView):
+
+    def get_title(self):
+        return _("Translate {model_name}").format(
+            model_name=capfirst(self.object._meta.verbose_name)
```
KalobTaulien, line 193: Playing devils advocate here: should all model names be capped in every instance? I've found that this covers _most_ instances but there are sometimes outliers where the model name is purposely lowercased all the time (when displaying model names)

wagtail_localize/wagtail_hooks.py
```diff
@@ -1,20 +1,63 @@
...
+
+
+@hooks.register('register_permissions')
+def register_submit_translation_permission():
+    return Permission.objects.filter(content_type__app_label='wagtail_localize', codename='submit_translation')
+
+
+@hooks.register("register_page_listing_more_buttons")
+def page_listing_more_buttons(page, page_perms, is_parent=False, next_url=None):
+    if page_perms.user.has_perms(['wagtail_localize.submit_translation']) and Locale.objects.exclude(id__in=page.get_translations(inclusive=True).values_list('locale_id', flat=True)).exists():
```
KalobTaulien, line 41: Could put on multiple lines. This is a pretty long conditional.

wagtail_localize/wagtail_hooks.py
```diff
@@ -1,20 +1,63 @@
...
+            url += '?' + urlencode({'next': next_url})
+
+        yield wagtailadmin_widgets.Button(_("Translate this page"), url, priority=60)
+
+
+@hooks.register('register_snippet_listing_buttons')
+def register_snippet_listing_buttons(snippet, user, next_url=None):
+    model = type(snippet)
+
+    if issubclass(model, TranslatableMixin) and user.has_perms(['wagtail_localize.submit_translation']) and Locale.objects.exclude(id__in=snippet.get_translations(inclusive=True).values_list('locale_id', flat=True)).exists():
```
KalobTaulien, line 53: Could put on multiple lines. This is a pretty long conditional, too.

wagtail_localize/tests/test_translation_model.py
kaedroho: That's correct. We don't actually use this value here

wagtail_localize/views/submit_translations.py
kaedroho: Yep, it does: https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L876

wagtail_localize/views/submit_translations.py
kaedroho: Tried changing this, but it broke the tests (validation wasn't running when it submitted the form without any locales selected)

wagtail_localize/templates/wagtail_localize/admin/submit_translation.html
kaedroho: Not thought much about browser support. Since this is a new plugin, do you think we need to support IE11?

wagtail_localize/views/submit_translations.py
kaedroho: Actually, yeah that makes sense. Probably only need to do this when we use the model name at the beginning of a sentence on its own.

wagtail_localize/templates/wagtail_localize/admin/submit_translation.html
kaedroho: Implemented a work around for this now!

PR #134, opened 2020-07-29
Initial frontend tooling

kaedroho:
Frontend tooling that gives us Typescript, SCSS and React. Taken from wagtail-review.

Planning to build the new edit UI and report in React.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+STATIC_SRC_DIR=wagtail_localize/static_src
+
+npm_deps:
+	cd ${STATIC_SRC_DIR} && npm install
+
+build: npm_deps
+	cd ${STATIC_SRC_DIR} && npm run build
+
+dev: npm_deps
+	cd ${STATIC_SRC_DIR} && npm run start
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -0,0 +1,10 @@
+STATIC_SRC_DIR=wagtail_localize/static_src
+
+npm_deps:
+	cd ${STATIC_SRC_DIR} && npm install
+
+build: npm_deps
+	cd ${STATIC_SRC_DIR} && npm run build
+
+dev: npm_deps
```
KalobTaulien, line 9: This will run `npm install` every time we want to watch files, won't it?

wagtail_localize/static_src/src/main.tsx
```diff
@@ -0,0 +1 @@
+// TODO: Add code here
```
KalobTaulien, line 1: TODO left behind

Makefile
kaedroho: Yep, it finishes really quickly when it hasn't got anything to do so I kinda like running it every time I start the watcher

wagtail_localize/static_src/src/main.tsx
kaedroho: Code coming in future PRs!

PR #146, opened 2020-08-05
Only allow one source per object/locale

kaedroho:
This PR changes the `TransLationSource` model to update instead of creating a new instance whenever the source object is updated.

This removes the ability for different translations to be at different versions, Mozilla doesn't want this but we can always add this back in later if we change our mind. Not having this feature simplifies some code and should save a good amount disk space over time!


wagtail_localize/migrations/0003_delete_translation_sources.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtail_localize', '0002_translation'),
+    ]
+
+    operations = [
+        migrations.RunPython(delete_translation_sources, migrations.RunPython.noop),
```
KalobTaulien, line 28: I'm fairly certain no, but does this data migration require a rewind function?

kaedroho: Iit just deletes stuff, so the rewind function would probably have to be to restore a backup!

kaedroho:
Thanks @KalobTaulien!

PR #159, opened 2020-08-10
Update translation view

kaedroho:
A view that lets you update existing translations for pages that are used as a source for any translations.

Any pages that are used as a source get a new item in the more menu "Update translations". This pushes all changes to that page to be translated into all the languages that page is currently translated into.

See it in action here (with older styling): https://localize-workflow-develop.herokuapp.com/admin/localize/update/1/?next=%2Fadmin%2Fpages%2F60%2F

### DIFF
--- a/wagtail_localize/views/update_translations.py
+++ b/wagtail_localize/views/update_translations.py
@@ -1,0 +1,109 @@
+from django import forms
+
+from django.contrib import messages
+from django.contrib.admin.utils import quote
+from django.core.exceptions import PermissionDenied
+from django.db import transaction
+from django.shortcuts import get_object_or_404, redirect
+from django.urls import reverse
+from django.utils.translation import gettext as _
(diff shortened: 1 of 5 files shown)


wagtail_localize/views/update_translations.py
```diff
@@ -0,0 +1,109 @@
...
+
+        if form.is_valid():
+            with transaction.atomic():
+                self.object.update_from_db()
+
+                if form.cleaned_data['publish_translations']:
+                    for translation in self.object.translations.select_related("target_locale"):
+                        translation.save_target(user=request.user, publish=True)
+
+                # TODO: Button that links to page in translations report when we have it
```
KalobTaulien, line 94: Remind on the TODO for this button

kaedroho: Will do this once I've added the report!

PR #164, opened 2020-08-20
Automatically generate translated_fields

kaedroho:
The rules around what should be translatable and what should be synchronised are fairly strict now, so we might as well automatically generate them to save the developer the hassle of learning all these rules!

I've left the manual approach available so our tests will continue to work


kaedroho:
TODO:
 - [ ] Detect from child relations

wagtail_localize/models.py
```diff
@@ -395,7 +396,7 @@ def create_or_update_translation(self, locale, user=None, publish=True, copy_par
             created = True
 
         # Copy synchronised fields
-        for field in getattr(translation, 'translatable_fields', []):
+        for field in get_translatable_fields(original.__class__):
             if field.is_synchronized(original):
                 # TODO: Use Django to set the field so the attname is correct
```
KalobTaulien, line 401: Is this TODO still required?

wagtail_localize/test/migrations/0003_nontranslatablechildobject_testgeneratetranslatablefieldspage_translatablechildobject.py
```diff
@@ -0,0 +1,70 @@
+# Generated by Django 3.0.8 on 2020-08-21 16:52
```
KalobTaulien, line 1: Greatest migration filename of all time.

kaedroho: Oh, yeah...

wagtail_localize/models.py
kaedroho: Yep, but it will go away in https://github.com/wagtail/wagtail-localize/pull/179

PR #182, opened 2020-09-05
Update travis base box to Ubuntu Focal Fossa

kaedroho:
To workaround SQLite bug:

This is it: https://code.djangoproject.com/ticket/24080
https://www.sqlite.org/src/info/7f7f8026eda38

This isn't a bug that affects localize. It's because our tests do a lot of rollbacks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,9 @@
+dist: focal
 language: python
 cache: pip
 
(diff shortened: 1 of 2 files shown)


kaedroho:
This is it: https://code.djangoproject.com/ticket/24080

PR #183, opened 2020-09-07
Display dependencies in editor

kaedroho:


wagtail_localize/views/edit_translation.py
```diff
@@ -407,7 +408,75 @@ def edit_translation(request, translation, instance):
...
+                'liveUrl': instance.full_url,
+                'editUrl': reverse('wagtailadmin_pages:edit', args=[instance.id]),
+                'createTranslationRequestUrl': reverse('wagtail_localize:submit_page_translation', args=[instance.id]),
+            }
+
+        else:
+            return {
+                'title': str(instance),
+                'isLive': True,
+                'editUrl': reverse('wagtailsnippets:edit', args=[instance._meta.app_label, instance._meta.model_name, quote(instance.id)]),
```
KalobTaulien, line 427: What if the model isnt registered as a snippet? Like if someone goes through the Django Admin, or through the Wagtail ModelAdmin?

KalobTaulien (approved):

wagtail_localize/views/edit_translation.py
kaedroho: We don't support django/modeladmin right now because it's really hard to tell if a model has been registered with either of those interfaces and what the user permissions are.

I think we should add a sort of "Model registry" into Wagtail, which we can query for this information (might be useful for API and generic choosers too).

kaedroho:
Thanks!

PR #209, opened 2020-10-06
Encode translation editor props with DjangoJSONEncoder

kaedroho:
Should hopefully fix those encoding errors when lazy translatable strings are used in field labels.


KalobTaulien (approved):

kaedroho:
Thanks!

PR #210, opened 2020-10-08
Update page draft_title immediately when title segment is edited

kaedroho:


wagtail_localize/views/edit_translation.py
```diff
@@ -283,6 +284,12 @@ def edit_translation(request, translation, instance):
                 for page in instance.get_ancestors(inclusive=False).descendant_of(cca, inclusive=True)
             ]
 
+        # Set to the ID of a string segment that represents the title.
+        # If this segment has a translation, the title will be replaced with that translation.
+        title_segment = string_segments.filter(context__path='title').first()
```
KalobTaulien, line 289: What happens if there's more than one item returned in the queryset? Will the .first() _always_ be the corrent title_segment to work with?

KalobTaulien (approved):

wagtail_localize/views/edit_translation.py
kaedroho: There should only be one segment per context, and one ``context__path`` per page. As all the segments must be for the same page, it should be unique.

I've changed this to a ``.get()`` just to be sure though!

PR #212, opened 2020-10-15
Drop support for Django 2.11 and 3.0

kaedroho:
I'd like to use the new generic JSONField in https://github.com/wagtail/wagtail-localize/pull/179

But also, since this is a new app, I'd like to drop these before the release so we don't have to deprecate them properly later!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,8 @@ services:
 
 matrix:
   include:
-   - env: TOXENV=py37-dj22-wamaster-postgres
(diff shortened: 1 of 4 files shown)


KalobTaulien (approved):
Good thinking on dropping old versions so there's less to deal with in terms of deprecation paths in the future. 

The minimum requirement now is Django 3.1 :tada:

kaedroho:
Just a note for anyone viewing this, version 1.0 will definitely support Django 2.11 and 3.0

kaedroho:
Postponing

PR #215, opened 2020-10-20
Remove Google Translate

kaedroho:
Not that comfortable with supporting this since it doesn't use the offical API.


KalobTaulien (approved):
Fair enough if it doesn't use the official API. Does it simply not work very well? And I'm curious if there's a way to give this feature to users who are mostly familiar with Google Translate (over translation services like DeepL) without having to support the non-api code. 

Either way, code looks good. 👍

kaedroho:
> Does it simply not work very well?

There's rate limits and it also doesn't support inline HTML.

The best way to provide this feature would be to add support for the offical API (but this costs money, like DeepL)

PR #243, opened 2020-11-12
Add override_translatable_fields option

kaedroho:
This option allows overriding specific fields but still translate other fields automatically.

For example:

```python
class MyCampaignPage(Page):
    campaign_id = models.CharField(max_length=255)

    override_translatable_fields = [
        # Prevent the campaign_id from being translated
        SynchronizedField('campaign_id'),
    ]
```


KalobTaulien (approved):

kaedroho:
Thanks!

PR #254, opened 2020-11-18
Extract fields that have the same copy separately

### LINKED ISSUE #253: When two fields have the same copy, content from only one is extracted
Just noticed that issue with the prod homepage of https://give.thunderbird.net, because staging had no `seo_title` set.

The `Page title` (or `seo_title`) field in the Promote tab has the following copy: `Give To Thunderbird`

The `Title` field in the Content tab (so, the actual page title) has the same copy, `Give To Thunderbird`

Only the seo_title got extracted in the .pot file: https://github.com/mozilla-l10n/thunderbird-donate-content/blob/master/templates/pages/mozilla-donate.pot#L8

We should have had something like this:
```
msgctxt "seo_title"
msgid "Give to Thunderbird"
msgstr ""

msgctxt "title"
msgid "Give to Thunderbird"
msgstr ""
```

kaedroho:
Fixes #253

### DIFF
--- a/wagtail_localize/models.py
+++ b/wagtail_localize/models.py
@@ -1,6 +1,5 @@
 import json
 import uuid
-from collections import defaultdict
 
 import polib
 from django.conf import settings
@@ -385,10 +384,10 @@ class TranslationSource(models.Model):
         Exports a PO file contining the source strings.
         """
         # Get messages
-        messages = defaultdict(list)
+        messages = []
 
         for string_segment in StringSegment.objects.filter(source=self).order_by('order').select_related("context", "string"):
-            messages[string_segment.string.data] = string_segment.context.path
+            messages.append((string_segment.string.data, string_segment.context.path))
 
         # Build a PO file
         po = polib.POFile(wrapwidth=200)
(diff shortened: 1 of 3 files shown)


wagtail_localize/models.py
```diff
@@ -363,10 +363,10 @@ def export_po(self):
         Exports a PO file contining the source strings.
         """
         # Get messages
-        messages = defaultdict(list)
+        messages = set()
```
KalobTaulien, line 366: Keep in mind a `set()` won't maintain its ordering.

KalobTaulien (approved):

wagtail_localize/models.py
kaedroho: Oh, I assumed they worked the same way as a dict! thanks for spotting, we definitely want that ordered

kaedroho: Fixed, I just turned it into a list. There isn't really a need to deduplicate things anyway!

TheoChevalier:
Hey @kaedroho, should this be just merged, or is there more work needed?

PR #257, opened 2020-11-20
Make TranslatableMixin for snippets not required

Vadim-Karpenko:
The snippets that do not have the TranslatableMixin or any translation features on it do not work due to an error message (see image below)
This update fixes the **"Nonetype" object has no attribute is_shown** error.
Here is an error message - 
image

### DIFF
--- a/wagtail_localize/wagtail_hooks.py
+++ b/wagtail_localize/wagtail_hooks.py
@@ -208,8 +208,15 @@ if SNIPPET_RESTART_TRANSLATION_ENABLED:
                 target_locale_id=context['instance'].locale_id,
                 enabled=False
             ).exists()
+    
+    
+    class DisabledTranslationSnippetActionMenuItem():
+        def is_shown(self, request, context):
+            return False
 
+    
     @hooks.register("register_snippet_action_menu_item")
     def register_restart_translation_snippet_action_menu_item(model):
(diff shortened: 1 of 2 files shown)


wagtail_localize/wagtail_hooks.py
```diff
@@ -208,8 +208,15 @@ def is_shown(self, request, context):
                 target_locale_id=context['instance'].locale_id,
                 enabled=False
             ).exists()
+    
+    
+    class DisabledTranslationSnippetActionMenuItem():
+        def is_shown(self, request, context):
```
kaedroho, line 214: Might be simpler to move the ``if issubclass(model, TranslatableMixin):`` check itself into the ``is_shown`` method on ``RestartTranslationSnippetActionMenuItem`` and then make the hook always return a ``RestartTranslationSnippetActionMenuItem``?

kaedroho (changes requested):
Thanks for fixing this! just a comment about a simpler approach that doesn't require a dummy menu item class. Also this needs a test! (test suite for this button is here: https://github.com/wagtail/wagtail-localize/blob/master/wagtail_localize/tests/test_submit_translations.py#L489)

kaedroho:
Fixes #265

kaedroho (approved):

PR #268, opened 2020-11-26
url patterns documentation

onno-timmerman:
url patterns documentation is using regex. This can confuse somebody … that installing it and sees more items in his setup with PATH.


kaedroho:
Thanks!

Issue #22, opened 2019-11-09
Translatable ListBlock


kaedroho:
We can't make `ListBlock` translatable because we need a way to uniquely identify blocks on different translations of a page while not being sensitive to ordering changes, deleting and adding new items.

The best alternative at the moment is to use a `StreamBlock` with a single block type, but the UI for this isn't very nice.

We can probably come up with a replacement for `ListBlock` that has the exact same UI and API, but the database representation is a list of objects with ``uuid`` and ``value`` keys.

kaedroho:
Just a note to say that we have a plan to improve ``ListBlock`` inside Wagtail itself to include an ID for each item

kaedroho:
RFC for adding UUIDs to `ListBlock` submitted here: https://github.com/wagtail/rfcs/pull/65

kaedroho:
Getting closer to a fix for this now: https://github.com/wagtail/wagtail/pull/7741

vladox:
@kaedroho do you have an estimation on when the PR will be released?

zerolab:
@vladox the aim is to get https://github.com/wagtail/wagtail/pull/7741/ in Wagtail 2.16, dependant on code review -- would help if you could test those changes with a local codebase.

Once it makes it in an official Wagtail release, we can proceed with this issue

PR #292, opened 2020-12-29
Localize list block

mardiros:
I've use a lot of ListBlock to avoid streamfield of one child type and now I discover that they are not translatable.

Here is the missing implementation.


kaedroho:
Thanks for looking into this!

Unfortunately, we can't use list block indexes in content paths as they can change between revisions, which would cause translations to be put in the wrong list block items.

There's a plan to fix this by adding UUIDs into `ListBlock` items in Wagtail core, see https://github.com/wagtail/wagtail-localize/issues/22

Until then, it's probably best to switch to `StreamBlock`

Issue #294, opened 2020-12-31
Adding invalid link identifier in translated block that does not exist in the source raise 500 errors while saving page


mardiros:
The following block should be marked as invalid and saving should be forbidden instead of raising the following traceback.

![image](https://user-images.githubusercontent.com/447396/103407040-4f55d900-4b5d-11eb-8781-82e14b176fa5.png)

```
[2020-12-31 10:40:26,248][624][ERROR][gn.home.management.commands.import_tld_translations] Error while processing skin
Traceback (most recent call last):
  File "/app/gn/home/management/commands/import_tld_translations.py", line 500, in handle
    tld_info, tld_page, locale, override=override, style=self.style,
  File "/app/gn/home/management/commands/import_tld_translations.py", line 464, in create_tld_translation_page
    translation.save_target(user, publish=True)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/models.py", line 863, in save_target
    self.source.create_or_update_translation(self.target_locale, user=user, publish=publish, fallback=True, copy_parent_pages=True)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/models.py", line 504, in create_or_update_translation
    ingest_segments(original, translation, self.locale, locale, segments)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 168, in ingest_segments
    ).handle_stream_block(data, field_segments)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 144, in handle_stream_block
    block.value = self.handle_block(block.block, block.value, segments)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 80, in handle_block
    return self.handle_struct_block(block_value, segments)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 112, in handle_struct_block
    block_type, block_value, segments
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 86, in handle_block
    return self.handle_stream_block(block_value, segments)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 144, in handle_stream_block
    block.value = self.handle_block(block.block, block.value, segments)
  File "/usr/local/lib/python3.7/site-packages/wagtail_localize/segments/ingest.py", line 80, in handle_block
[...]

PR #295, opened 2020-12-31
Check for invaid id on <a> while submitting translations

mardiros:
Adding this request to fix the issues I've open this morning:

https://github.com/wagtail/wagtail-localize/issues/294

Now, when I have an error that flash:

image

Unfortunatly, because the error is saved in the database, existing errors that are not marked in the database cannot
be found, but I this it is ok like that. I guess error are saved for performance reason and it is fine for me.

Hope its helps :smiley:

### DIFF
--- a/wagtail_localize/strings.py
+++ b/wagtail_localize/strings.py
@@ -389,3 +389,28 @@ def restore_strings(template, strings):
         text_element.replaceWith(string.render_soup(attrs))
 
     return str(soup)
+
+
+
+def extract_ids(template):
+    """Extract link ids from one template string and return it in a set."""
+    soup = BeautifulSoup(template, "html.parser")
+    ids = set()
+    for element in soup.descendants:
+        if not isinstance(element, Tag):
+            continue
+
+        if element.name == 'a':
+            if 'id' in element.attrs:
+                ids.add(element.attrs['id'])
+
+    return ids
+
+
+def validate_translation_links(translation_of, data):
+    """Check that the link id in a translation are present in its source."""
+    id1s, id2s = extract_ids(translation_of), extract_ids(data)
+    new_ids = id2s - id1s
+    if new_ids:
+        ids = ", ".join(sorted(new_ids))
+        raise ValueError(_("Unrecognised id found in an <a> tag: {}").format(ids))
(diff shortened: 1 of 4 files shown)


mardiros:
cc @kaedroho

wagtail_localize/strings.py
```diff
@@ -389,3 +389,28 @@ def restore_strings(template, strings):
...
+
+def extract_ids(template):
+    """Extract link ids from one template string and return it in a set."""
+    soup = BeautifulSoup(template, "html.parser")
+    ids = set()
+    for element in soup.descendants:
+        if not isinstance(element, Tag):
+            continue
+
+        if element.name == 'a':
```
kaedroho, line 403: I think we could get rid of this check, technically any inline tag with attributes will be given an `id` (like custom inline tags for example)

wagtail_localize/strings.py
```diff
@@ -389,3 +389,28 @@ def restore_strings(template, strings):
...
+    return ids
+
+
+def validate_translation_links(translation_of, data):
+    """Check that the link id in a translation are present in its source."""
+    id1s, id2s = extract_ids(translation_of), extract_ids(data)
+    new_ids = id2s - id1s
+    if new_ids:
+        ids = ", ".join(sorted(new_ids))
+        raise ValueError(_("Invalid id found in <a>: {}").format(ids))
```
kaedroho, line 416: "Unrecognised" is probably more fitting than "Invalid" here. Also, as above, this could happen in non-a tags too

kaedroho (commented):
Thanks, this looks good! Only thing I could spot is it should work on other inline tags and also a wording tweak

wagtail_localize/strings.py
mardiros: Do you want me to remove that ?

I've made it because currently there is

```
        # Elements can't have attributes, except for <a> tags
        keys = set(element.attrs.keys())
        if element.name == 'a' and 'id' in keys:
            keys.remove('id')
        if keys:
            raise ValueError(_("Strings cannot have any HTML tags with attributes (except for 'id' in <a> tags)"))
```

kaedroho: Oh, sorry I forgot I added that! Yep, this is fine how it is, we can deal with custom tags later

wagtail_localize/strings.py
kaedroho: Sorry just noticed this one, could we add more wording around the <a>: "found in an <a> tag"

mardiros: Ok, too many commits for that small fix, now it is `"Unrecognised id found in an <a> tag: {}"`

my last commit of the year :champagne:

kaedroho (approved):
Looks good, thanks for fixing!

PR #296, opened 2021-01-04
Extract links to segment and restore them while translating

mardiros:
@kaedroho 

Currently there is no way to translate external links in richtext because they are out of the segment, saved and restored.

but we need to translate links for exemple https://12factor.net/ to https://12factor.net/fr/

I propose here a patch to make them translatable.

If you are ok with the approach, I can add tests and screenshot here.

### DIFF
--- a/wagtail_localize/tests/test_edit_translation.py
+++ b/wagtail_localize/tests/test_edit_translation.py
@@ -184,6 +184,7 @@ class TestGetEditTranslationView(EditTranslationTestData, TestCase):
         self.assertEqual(
             [(segment['contentPath'], segment['value']) for segment in props['segments'] if segment['type'] == 'synchronised_value'],
             [
+                ('test_richtextfield.http://example.com', 'http://example.com'),
                 ('test_synchronized_emailfield', 'email@example.com'),
             ]
         )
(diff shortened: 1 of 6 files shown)


kaedroho:
Thanks!

It looks like it's extracting links as translatable strings, but I think we should extract them as overridable values so that they don't get put in PO files, or machine translated, but they would be changeable in the admin. Does that would OK for you?

I think we could implement this by checking the `attrs`` for links [here](https://github.com/wagtail/wagtail-localize/blob/master/wagtail_localize/segments/extract.py#L48) and [here](https://github.com/wagtail/wagtail-localize/blob/master/wagtail_localize/segments/extract.py#L140) and add an ``OverridableSegmentValue`` to the segments list for each link it finds (append the id of the tag it came from to the segment path, so we know where to ingest it later).

On ingestion, we could check if any ``OverridableSegmentValue`` segments were passed in to the [``organise_template_segments`` function](https://github.com/wagtail/wagtail-localize/blob/master/wagtail_localize/segments/ingest.py#L15) and update the ``attrs`` of the relevant string(s) if so.

mardiros:
@kaedroho 

I try the way you want, I did not cleanup all the things actually.

What I dislike is that organise_template_segments is going to update things and the function name looks to prepare data
to be translated, I don't know if it is a problem for you.

Then, I have another issue I've tried to solve and I try all the afternoon but I stuck.

I have fields without the edit button,
image

because they have the type "unknown"

https://github.com/wagtail/wagtail-localize/pull/296/commits/cd8838a120e95675208353b01bc4a736bdc3f838

This is because the field is a streamfield in a streamfield

and the funciton {get_segment_location_info](https://github.com/wagtail/wagtail-localize/blob/fdd813e8a9717b8336f31ba9d04b75976d1d915f/wagtail_localize/views/edit_translation.py#L297) is looking at the first streamfield and it did not kwown the type.

I will revert that fix, I've commit to let you know that. Should I open another issue for that ?

mardiros:
@kaedroho 

I've cleanup the code now, to avoid temporary `xref` attrs, and use OverridableSegment for links.

I've also remove the patch that are not related to this feature.

I will update the test suite if you are OK with this implementation, which is my comprehension of what

your comment https://github.com/wagtail/wagtail-localize/pull/296#issuecomment-754151854

wagtail_localize/tests/test_edit_translation.py
```diff
@@ -184,19 +184,27 @@ def test_edit_page_translation(self):
         self.assertEqual(
             [(segment['contentPath'], segment['value']) for segment in props['segments'] if segment['type'] == 'synchronised_value'],
             [
+                ('test_richtextfield.http://example.com', 'http://example.com'),
```
kaedroho, line 187: To future-proof this, we should add some syntax that makes it clear that the ``.`` in ``example.com`` isn't a content path separator. Perhaps we should quote values like this? For example: ``test_richtextfield."http://example.com"``

kaedroho:
Thanks!

mardiros:
Many thanks for the merge !

PR #357, opened 2021-03-01
Ensure we remove translation of source that have been removed

mardiros:
Hello @kaedroho 

We have currently a bug in the translation.

The flow is:

 we have a field named "strapline", and we have to clean it in some pages.

 * in the translation editor, the source disapear

 * after publishing, the translation is still there because it has not been cleaned in the translated page.

All tables (wagtail_localized_string, wagtail_localized_stringtranslation, ...) are cleaned but the data is still in the published page.

Here is my pull request for the fix.

Tell me if you need more explanation of the bug.

### DIFF
--- a/wagtail_localize/segments/ingest.py
+++ b/wagtail_localize/segments/ingest.py
@@ -8,6 +8,7 @@ from wagtail.core.fields import RichTextField, StreamField
 from wagtail.core.rich_text import RichText
 
 from wagtail_localize.strings import restore_strings
+from wagtail_localize.fields import get_translatable_fields
 
 from .types import OverridableSegmentValue, StringSegmentValue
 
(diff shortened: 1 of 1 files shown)


wagtail_localize/segments/ingest.py
```diff
@@ -226,10 +227,23 @@ def ingest_segments(original_obj, translated_obj, src_locale, tgt_locale, segmen
...
-        if hasattr(field, "restore_translated_segments"):
+        if not field_segments:
+
+            if isinstance(field, (models.TextField, models.CharField)):
+                data = field.value_from_object(original_obj)
+            else:
+                # TODO implement for other types
+                continue
+
+            setattr(translated_obj, field_name, data)
```
kaedroho, line 244: To synchronise a field with the source, you should register it as a ``SynchronisedField``. I don't think localize should try to be clever about fields that are not configured

wagtail_localize/segments/ingest.py
```diff
@@ -226,10 +227,23 @@ def ingest_segments(original_obj, translated_obj, src_locale, tgt_locale, segmen
         field_name, segment = segment.unwrap()
         segments_by_field_name[field_name].append(segment)
 
-    for field_name, field_segments in segments_by_field_name.items():
+    for field in get_translatable_fields(translated_obj.__class__):
```
kaedroho, line 230: Makes sense, can't remember why I didn't do this before!

wagtail_localize/segments/ingest.py
```diff
@@ -226,10 +227,23 @@ def ingest_segments(original_obj, translated_obj, src_locale, tgt_locale, segmen
...
+
+        field_name = field.field_name
+        field_segments = segments_by_field_name[field_name]
         field = translated_obj.__class__._meta.get_field(field_name)
 
-        if hasattr(field, "restore_translated_segments"):
+        if not field_segments:
+
+            if isinstance(field, (models.TextField, models.CharField)):
```
kaedroho, line 238: Maybe we should synchronise all translatable fields before restoring translations? We already do this for streamfield

zerolab:
@mardiros I believe this is now handled by #622 / #635 and included in v1.3.x
I realise it took some time, but would appreciate a check on your side

mardiros:
Thanks @zerolab 

I close this pull request, and will open an issue if something will go wrong on my next update.

PR #366, opened 2021-03-26
Update locale files

imomaliev:


kaedroho:
Thanks! I've just done this and pushed to transifex

PR #399, opened 2021-06-01
Add CI workflow task to detect new strings

kaedroho:
This PR adds a new Github workflow task that checks if there are any changes to translatable strings that haven't yet been extracted into a PO file.

I've set up a one-way link from this repo to Transifex so that translations are updated there as soon as we've generated the PO files here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,3 +88,39 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           fail_ci_if_error: true
+
+  new-strings:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: ['3.8']
+        node: ['14']
+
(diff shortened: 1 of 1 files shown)


KalobTaulien (approved):

kaedroho:
@zerolab 

This PR adds a Github actions check in to alert us if we merge anything that requires the PO files to be rebuilt.

The reason why I haven't merged it is just that I thought it needed to display a more useful message of the check failed.

Happy to add this in if you think this is still worth having.

zerolab:
This is certainly useful. I agree that it could use a more meaningful message.
AWill this fail if there are new messages? so it is more than just an info blurb?

zerolab:
I think I will close this now as the workflow we had so far worked well in my opinion.
Thank you for this (and all the other PRs) Karl

PR #416, opened 2021-06-15
Use gettext_lazy to avoid generating unnecessary migrations

vparitskiy:
When your LANGUAGE_CODE is not `en` and strings `Manual` or `Machine` have already been added to translations, usage of non lazy gettext generates new strings inside the `makemigrations` runtime, which leads to Django false positively detecting changes in models.


kaedroho:
Thanks!

PR #439, opened 2021-07-24
Support overridable ForeignKey pointing to UUID primary key

hpoul:
I stumbled on an error when translating a Page model which had a `ForeignKey` to another `Page` which had a primary key of type `UUID`, because it couldn't be serialized.
Since most `json.dumps` methods already use the `DjangoJSONEncoder` it seems logical to me that this should be included for `OverridableSegment` as well?


kaedroho (changes requested):
This change looks good to me, thanks! Please could you add a test?

hpoul:
thanks for taking a look @kaedroho - unfortunately I'm still getting to know the internals of wagtail and wagtail-localize, maybe you could give me some pointers of existing tests which would call that method.. my idea was to basically copy https://github.com/wagtail/wagtail-localize/blob/f5880059cc84ee96aab5ade17f78dc6112b04462/wagtail_localize/tests/test_translation_model.py#L513-L532 `test_save_target_snippet` which sounded like an easy test which would translate a snippet..

unfortunately i ran into a weird problem.. i've got the following model:

```
@register_snippet
class TestUUIDSnippet(TranslatableMixin, models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    field = models.TextField(gettext_lazy("field"))

    translatable_fields = [TranslatableField("field")]
```

and completely the same test just with that UUID snippet would throw me an error like:

```
System check identified 10 issues (0 silenced).
......E
======================================================================
ERROR: test_uuid_snippet_save_target (wagtail_localize.tests.test_translation_model.TestSaveTarget)
----------------------------------------------------------------------
  File "/Users/herbert/dev/python/wagtail-localize/wagtail_localize/tests/test_translation_model.py", line 570, in test_uuid_snippet_save_target
    translation.save_target()
  File "/Users/herbert/dev/python/wagtail-localize/wagtail_localize/models.py", line 1134, in save_target
    self.source.create_or_update_translation(self.target_locale, user=user, publish=publish, fallback=True, copy_parent_pages=True)
  File "/Users/herbert/dev/python/wagtail-localize/wagtail_localize/models.py", line 727, in create_or_update_translation
    translation.full_clean()
  File "/Users/herbert/dev/python/wagtail-localize/venv/lib/python3.9/site-packages/django/db/models/base.py", line 1238, in full_clean
    raise ValidationError(errors)
django.core.exceptions.ValidationError: {'id': ['Test uuid snippet with this Id already exists.']}

----------------------------------------------------------------------
Ran 7 tests in 0.912s
```

Maybe you could let me know if i'm at least on the right track with that test, I'm not completely sure when a `OverridableSegment` would be created programmatically 🤔️

hpoul:
Okay, I revisited the problem and I made the test wrong. I needed a snippet which contained a `ForeignKey` to a model with a `UUIDField`. The added test fails without the change, but succeeds with the change.
@kaedroho maybe you could take another look :-)

wagtail_localize/test/migrations/0021_testuuidmodel_testuuidsnippet.py
```diff
@@ -0,0 +1,72 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("wagtailcore", "0062_comment_models_and_pagesubscription"),
```
zerolab, line 11: I'd say drop this line as it is the source of test failures

zerolab (changes requested):
One fix for the test failures needed. Otherwise LGTM

wagtail_localize/test/migrations/0021_testuuidmodel_testuuidsnippet.py
hpoul: thanks.. removed it.. let's hope it fixed things 🤞

hpoul:
@zerolab sorry, i guess there was a merge conflict, I've now rebased on master.. tests work locally now, i guess 🤔

```
(venv) ➜  wagtail-localize git:(foreignkey-uuid) tox -e python3.9-django4.0-wagtail2.16-sqlite
[...]
Ran 331 tests in 79.061s

OK (expected failures=4)
Destroying test database for alias 'default' ('file:memorydb_default?mode=memory&cache=shared')...
```

zerolab (approved):

wagtail_localize/test/models.py
```diff
@@ -43,6 +45,19 @@ class TestSnippet(TranslatableMixin, models.Model):
...
 
 
+class TestUUIDModel(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    charfield = models.CharField(max_length=10, blank=True)
+
+
+@register_snippet
+class TestUUIDSnippet(TranslatableMixin, models.Model):
+    # id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
chris48s, line 55: Stray bit of commented out code here. If it doesn't need to be here lets drop it.

chris48s (commented):
One very minor comment. Aside from that LGTM

chris48s (approved):

zerolab:
Nice one, thank you @hpoul for the PR and @chris48s for the review

PR #473, opened 2021-10-21
Add a means to disable the default translation mode

### LINKED ISSUE #447: Add option to create translations with "translation mode" off by default
Currently, whenever you create translations for a page, you are always in "translation mode" by default. This may usually be what you want, but translation mode has some limitations and in some cases you might want to always go straight to creating fully localized versions of a page based on a copy of the page, rather than using the translation mode. In these cases it might be nice to be able to change the default so that a translated page starts out with "translation mode" off.

I'd suggest that the following approach would be most flexible:
* A global setting that impacts ALL content in case you always want to skip translation mode
* And, a property you can set per page type, in case you want most pages to use translation mode but there are specific more complex pages you would prefer to skip translation mode on.

zerolab:
Fixes #447

This PR adds a new global `WAGTAIL_LOCALIZE_DEFAULT_TRANSLATION_MODE` setting that allows a developer to disable the default translation editor by setting this value to ``'simple'``

It also checks for a `localize_default_translation_mode` attribute on the source model allowing for more specificity (e.g. globally disabled translation mode, but enabled for a particular page type. Or the other way around)


docs/how-to/installation.md
```diff
@@ -34,3 +34,18 @@ The locales that are synchronised from others can be managed through the Wagtail
 
 To enable this, replace the ``"wagtail.locales"`` entry in ``INSTALLED_APPS`` with ``"wagtail_localize.locales"``, this
 will add a "Sync from" field to all locales that allows an administrator to choose a locale to synchronise content from.
+
+
+## Disabling the default translation mode
+
+Wagtail Localize will automatically go in translation mode when creating new translations for models.
+
+To disable this behaviour globally, set `WAGTAIL_LOCALIZE_DISABLE_TRANSLATION_MODE = True` in your settings file.
```
kaedroho, line 43: Wondering if "disable" is the best wording here? As it doesn't prevent the user from using any translation modes

How about ``WAGTAIL_LOCALIZE_DEFAULT_TRANSLATION_MODE = 'synced' / 'simple'``?

zerolab: I went in circles a few times. Happy to change this

PR #481, opened 2021-11-02
Specify Postgres version for the Wagtail main env test in CI

zerolab:
Fixes this https://github.com/wagtail/wagtail-localize/runs/4070369548?check_suite_focus=true CI failure

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
         "Framework :: Wagtail",
         "Framework :: Wagtail :: 2",
     ],
-    install_requires=["Django>=2.2,<4.0", "Wagtail>=2.11,<2.16", "polib>=1.1,<2.0"],
+    install_requires=["Django>=2.2,<4.0", "Wagtail>=2.11,<2.17", "polib>=1.1,<2.0"],
     extras_require={
         "testing": ["dj-database-url==0.5.0", "freezegun==0.3.15"],
         "documentation": [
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@
         "Framework :: Wagtail",
         "Framework :: Wagtail :: 2",
     ],
-    install_requires=["Django>=2.2,<4.0", "Wagtail>=2.11,<2.16", "polib>=1.1,<2.0"],
+    install_requires=["Django>=2.2,<4.0", "Wagtail>=2.11,<2.17", "polib>=1.1,<2.0"],
```
zerolab, line 44: This is probably the more controversial change. In order to run the tests agains Wagtail main, we want the upper bound to be vNext+1

kaedroho: I think we should avoid doing this here since there's a good chance we could accidentally ship a package that falsely advertises 2.16 support and cause issues later.

zerolab: any suggestion on how to work around https://github.com/wagtail/wagtail-localize/runs/4079597377?check_suite_focus=true

```
ERROR: Cannot install wagtail 2.16a0 (from git+https://github.com/wagtail/wagtail.git) and wagtail-localize[testing]==1.0rc4 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested wagtail 2.16a0 (from git+https://github.com/wagtail/wagtail.git)
    wagtail-localize[testing] 1.0rc4 depends on Wagtail<2.16 and >=2.11
```

zerolab:
Fixed in #509

PR #488, opened 2021-11-09
Lint docstrings

kaedroho:
This PR adds flake8-docstrings and lints all the existing docstrings.

We haven't used this flake8 plugin before, so just trying it out.

What do you think?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
       - image: circleci/python:3.8
     steps:
       - checkout
-      - run: pip install flake8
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       - image: circleci/python:3.8
     steps:
       - checkout
-      - run: pip install flake8
+      - run: pip install flake8 flake8-docstrings
```
zerolab, line 9: It is an interesting idea and I don't see the harm in it.

Would be good to add to the pre-commit config as well

PR #491, opened 2021-11-22
Translation submit/update form components

zerolab:
This PR provides a mechanism to add new form components to the submit/update translations views.

It takes the existing mechanism used on the locale edit forms

When submitting a translation we can get one or more translation sources with at least one corresponding translation sources. This should be accounted for by hte new `TranslationComponentManager`.

New component registration can be done using the new `register_translation_component` decorator

```python
@register_translation_component(
    heading=gettext_lazy("My component"),
    help_text=gettext_lazy("Help text to describe what this is needed for"),
    enable_text=gettext_lazy("Add custom component data"),
    disable_text=gettext_lazy("Do not add custom component data"),
)
class MyTranslationComponent(model.Model):
    pass
```

Did not yet add documentation as `register_locale_component` is not documented. Let's discuss


kaedroho (approved):
This all looks good to me, nice work!

PR #495, opened 2021-12-13
Handle text field clearing when syncing

### LINKED ISSUE #467: Frontend on synchronised pages still shows deleted fields
The frontend of a page that is being translated from another locale (in synchronised mode) still shows field that it shoudn't sometimes. Specifically this is when you delete all text from a field on the source page. To reproduce:

1. Go to a source page and delete all text in a text field
2. Publish that page
3. Synchronise that page to another locale
4. Go to the other locale's page and check the frontend. The text will still show. (the field won't show in the admin at all though)

zerolab:
Fixes #467

tl;dr - we mark text fields and `StreamField`s as `TranslateableField`, but only flag `StreamField`s as needing syncing. When the value is removed, there is nothing to tell `Translation.create_or_update_translation()` that this has happened.

`Translation._get_segments_for_translation()` only return translateable fields that have values in them, so nothing will happen via `ingest_segments`

### DIFF
--- a/wagtail_localize/fields.py
+++ b/wagtail_localize/fields.py
@@ -61,16 +61,24 @@ class TranslatableField(BaseTranslatableField):
     def is_translated(self, obj):
         return True
 
-    def is_synchronized(self, obj):
-        field = self.get_field(obj.__class__)
+    def is_synchronized(self, source):
+        field = self.get_field(source.__class__)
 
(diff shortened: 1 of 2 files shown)


chris48s (approved):
I don't think my :+1: counts for anything on this repo but LGTM

PR #496, opened 2021-12-14
Sync source page privacy settings with translated page

### LINKED ISSUE #317: Translations are not inheriting the source page privacy settings
This is the same issue as https://github.com/wagtail/wagtail/issues/6750, but for translated pages.

zerolab:
This PR fixes #317

it synchronizes the source page `PageViewRestriction` with the translated pages on translation creation and sync.

Scenarios covered:

Source | Translation | Action
-------|-------------|-----------
Has privacy settings | Doesn't have privacy settings | Translated page privacy settings get created
Has changes to the privacy settings | Has privacy settings | Translated page privacy settings get updated
Has privacy settings removed | Has privacy settings | Translated page privacy settings removed
Has privacy settings removed | Doesn't have privacy settings | Noop

Note: if the source page privacy settings change, the translated page(s) privacy settings will only get updated after "sync translations"

### DIFF
--- a/wagtail_localize/models.py
+++ b/wagtail_localize/models.py
@@ -234,6 +234,14 @@ class CannotSaveDraftError(Exception):
     pass
 
 
+class NoViewRestrictionsError(Exception):
+    """
+    Raised when trying to sync view restrictions for non-Page objects
+    """
+
+    pass
+
+
 class MissingTranslationError(Exception):
     def __init__(self, segment, locale):
         self.segment = segment
@@ -888,7 +896,7 @@ class TranslationSource(models.Model):
             translation_page (Page|Snippet): The translated instance.
         """
         if not isinstance(original, Page) or not isinstance(translation_page, Page):
-            return
+            raise NoViewRestrictionsError
 
         if original.view_restrictions.exists():
             original_restriction = original.view_restrictions.first()
@@ -943,7 +951,7 @@ class TranslationSource(models.Model):
 
         try:
             translation_page = self.get_translated_instance(locale)
-        except models.ObjectDoesNotExist:
+        except Page.DoesNotExist:
             return
 
         self.sync_view_restrictions(original, translation_page)
--- a/wagtail_localize/models.py
+++ b/wagtail_localize/models.py
@@ -42,6 +42,7 @@ from wagtail.core.models import (
     Page,
     PageLogEntry,
     TranslatableMixin,
+    _copy,
     get_translatable_models,
 )
 from wagtail.core.utils import find_available_slug
@@ -772,6 +773,8 @@ class TranslationSource(models.Model):
                     # Create a new revision
                     page_revision = translation.save_revision(user=user)
 
+                    self.sync_view_restrictions(original, translation)
+
                     if publish:
                         page_revision.publish()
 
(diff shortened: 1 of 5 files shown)


wagtail_localize/models.py
```diff
@@ -876,6 +879,75 @@ def schema_out_of_date(self):
...
+            original (Page|Snippet): The original instance.
+            translation_page (Page|Snippet): The translated instance.
+        """
+        if not isinstance(original, Page) or not isinstance(translation_page, Page):
+            return
+
+        if original.view_restrictions.exists():
+            original_restriction = original.view_restrictions.first()
+            if not translation_page.view_restrictions.exists():
+                view_restriction, child_object_map = _copy(
```
chris48s, line 896: Minor style point but `child_object_map` is an used variable here

Normally I'd suggest

```py
view_restriction, _ = _copy(...
```

but we've done `from django.utils.translation import gettext as _` in this file, so probably not that :)

If we wanted to avoid the unused var we could do 


```py
view_restriction = _copy(
    original_restriction,
    exclude_fields=["id"],
    update_attrs={"page": translation_page}
)[0]
```

I think I'll just flag it and leave you to decide what you think is most readable though

wagtail_localize/models.py
```diff
@@ -876,6 +879,75 @@ def schema_out_of_date(self):
...
+                    should_save = True
+                if translation_restriction.password != original_restriction.password:
+                    translation_restriction.password = original_restriction.password
+                    should_save = True
+                if list(
+                    original_restriction.groups.values_list("pk", flat=True)
+                ) != list(translation_restriction.groups.values_list("pk", flat=True)):
+                    translation_restriction.groups.set(
+                        original_restriction.groups.all()
+                    )
```
chris48s, line 922: Are wemissing

`should_save = True`

in this last if block?

chris48s: Also, just for my own understanding, is the reason why we're doing this

```py
should_save = False
if this != that:
    this = that
    should_save = True

if should_save:
    this.save()
```

pattern rather than just

```py
this = that
this.save()
```

for performance reasons? (i.e: we're trying to minimise round trips to the DB)

wagtail_localize/models.py
zerolab: definitely not `_` because of the `gettext` import

we have a number of other places where this is happening, so I follow it. the `[0]` variant will work, but it is not the most readabale so I think we can live with the unused var

wagtail_localize/models.py
zerolab: it certainly is missing `should_save = True`.

The reason for this is that 2 or more of the conditions can be true.

e.g. when changing form the 'login' restriction to 'password', both the `restriction_type` and `password` attributes will have changed.

so we want a single save

zerolab: on missing `should_save`, we are not - https://github.com/wagtail/wagtail-localize/pull/496/commits/0741ae45233ed06f11a65d35c6995d31e6ff75e2#diff-a9c40e8134c27fd95a23649e865005e8a957dc24db0d423122655a565221d614R555-R565 - as groups are M2M

chris48s (approved):

PR #500, opened 2022-01-05
Fix default REST_FRAMEWORK permission classes could break api views (#499)

### LINKED ISSUE #499: edit_string_translation view should not use REST_FRAMEWORK default permissions and authentication classes
[edit_string_translation](https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/views/edit_translation.py#L986)  api view currently does not explicitly specify permission class, which means that default settings for rest framework would be used. 

When [DjangoModelPermissionsOrAnonReadOnly](https://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissionsoranonreadonly) is used as a default permission class:

```python
REST_FRAMEWORK = {
    'DEFAULT_AUTHENTICATION_CLASSES': [
        'rest_framework.authentication.SessionAuthentication',
    ],
    # Use Django's standard `django.contrib.auth` permissions,
    # or allow read-only access for unauthenticated users.
    "DEFAULT_PERMISSION_CLASSES": [
        "rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly"
    ],
}
```

following error would be raised in ajax view to translate string:

```
Exception Type: AssertionError at /cms/localize/translate/1/strings/1/edit/
Exception Value: Cannot apply DjangoModelPermissionsOrAnonReadOnly on a view that does not set .queryset or have a .get_queryset() method.
```

I think that `edit_string_translation` view function should be decorated with required permission class to avoid taking defaults, ie:

```python
@permission_classes([IsAuthenticated])
```

bmihelac:
Fixes #499

Unfortunately, it seems there is no easy way to test this. DRF `api_views` decorator apply default permission classes early, and `settings_override` has no effect. I have added `REST_FRAMEWORK` settings that are known to break `wagtail_localize` in test app settings.

### DIFF
--- a/wagtail_localize/test/settings.py
+++ b/wagtail_localize/test/settings.py
@@ -179,3 +179,12 @@ WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
 
 
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+
+
+# wagtail localize should not use DEFAULT_PERMISSION_CLASSES
+# see: https://github.com/wagtail/wagtail-localize/issues/499
+REST_FRAMEWORK = {
+    "DEFAULT_PERMISSION_CLASSES": [
+        "rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly"
+    ],
+}
--- a/wagtail_localize/test/settings.py
+++ b/wagtail_localize/test/settings.py
@@ -183,8 +183,8 @@ DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 # wagtail localize should not use DEFAULT_PERMISSION_CLASSES
 # see: https://github.com/wagtail/wagtail-localize/issues/499
-REST_FRAMEWORK = {
-    "DEFAULT_PERMISSION_CLASSES": [
-        "rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly"
-    ],
-}
+# REST_FRAMEWORK = {
+    # "DEFAULT_PERMISSION_CLASSES": [
+        # "rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly"
+    # ],
+# }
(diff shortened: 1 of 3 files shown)


zerolab:
@bmihelac according to https://github.com/encode/django-rest-framework/blob/master/tests/test_settings.py#L29-L44 you should be able to use `override_settings`

bmihelac:
@zerolab I have tried and it does not work. 

I have added test code in https://github.com/wagtail/wagtail-localize/pull/500/commits/d6ec08bc51602c2a0e896462348c77fadaebad2e comiit.

It can be checked with

```
python testmanage.py test wagtail_localize.tests.test_edit_translation.TestEditOverrideAPIView.test_update_override_with_default_permissions_classes
```

`api_settings.DEFAULT_PERMISSION_CLASSES` is updated and it correctly shows `[<class 'rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly'>]` but view still have `rest_framework.permissions.AllowAny`. I guess that is because view is decorated early in:

https://github.com/encode/django-rest-framework/blob/a780e80debac0d12bb62fbb0ed98635e601e76de/rest_framework/decorators.py#L70-L71

bmihelac:
@zerolab if there is any other idea how tests for this could be added, I would be happy to update this PR

zerolab:
@bmihelac will have a look over the weekend and get back to you

zerolab:
@bmihelac this is on my list for tomorrow. Sorry for the delay

bmihelac:
No problems @zerolab 
iI you think there is anything that I could try let me know, I am out of ideas other than opening issue on DRF.

zerolab:
@bmihelac opened #513 which fixes the tests as well as addresses a related issue (having to do with default permission classes not including `SessionAuthentication`)

Will close this. Thank you very much for your PR!

PR #505, opened 2022-01-18
Support Deepl free api

vladox:
Essentially it enables the Deepl API to be changed since they have two different (free and pro).
Also some improvements were made on error handling:

image

There's a new option to define the Deepl API endpoint:

```
WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
    "CLASS": "wagtail_localize.machine_translators.deepl.DeepLTranslator",
    "OPTIONS": {
        "AUTH_KEY": os.environ.get('DEEPL_API_KEY', None),
        "API_ENDPOINT": os.environ.get('DEEPL_API_ENDPOINT', None),
    },
}
```

### DIFF
--- a/wagtail_localize/machine_translators/deepl.py
+++ b/wagtail_localize/machine_translators/deepl.py
@@ -1,8 +1,18 @@
 import requests
 
+from http import HTTPStatus
+
 from wagtail_localize.strings import StringValue
 
 from .base import BaseMachineTranslator
+
+class ApiEndpointError(Exception):
+
+    def __init__(self, code, reason, message="Unspecified Deepl API error ocurred"):
+        self.code = code
+        self.reason = reason
+        self.message = message
+        super().__init__(f"{self.code}: [{self.reason}] {self.message}")
 
 
 def language_code(code, is_target=False):
@@ -17,8 +27,11 @@ class DeepLTranslator(BaseMachineTranslator):
     display_name = "DeepL"
 
     def translate(self, source_locale, target_locale, strings):
+        api_endpoint = "https://api.deepl.com/v2/translate"
+        if "API_ENDPOINT" in self.options:
+            api_endpoint = self.options["API_ENDPOINT"]
         response = requests.post(
-            "https://api.deepl.com/v2/translate",
+            api_endpoint,
             {
                 "auth_key": self.options["AUTH_KEY"],
                 "text": [string.data for string in strings],
@@ -29,6 +42,9 @@ class DeepLTranslator(BaseMachineTranslator):
                 ),
             },
         )
+
+        if response.status_code != HTTPStatus.OK:
+            raise ApiEndpointError(response.code, response.reason, response.text)
 
         return {
             string: StringValue(translation["text"])
--- a/wagtail_localize/machine_translators/deepl.py
+++ b/wagtail_localize/machine_translators/deepl.py
@@ -44,7 +44,7 @@ class DeepLTranslator(BaseMachineTranslator):
         )
 
         if response.status_code != HTTPStatus.OK:
-            raise ApiEndpointError(response.code, response.reason, response.text)
+            raise ApiEndpointError(response.status_code, response.reason, response.text)
 
         return {
             string: StringValue(translation["text"])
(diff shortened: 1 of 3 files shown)


wagtail_localize/machine_translators/deepl.py
```diff
@@ -17,8 +27,11 @@ class DeepLTranslator(BaseMachineTranslator):
     display_name = "DeepL"
 
     def translate(self, source_locale, target_locale, strings):
+        api_endpoint = "https://api.deepl.com/v2/translate"
+        if "API_ENDPOINT" in self.options:
+            api_endpoint = self.options["API_ENDPOINT"]
```
zerolab, line 32: suggestion:

```suggestion
        api_endpoint = self.options.get("API_ENDPOINT", "https://api.deepl.com/v2/translate")
```

zerolab (changes requested):
Thanks you @vladox!
Did a quick first pass review with a minor suggestion.
The main thing here is we need a couple of tests (one for the setting and the other for the error message)

wagtail_localize/machine_translators/deepl.py
vladox: @zerolab thanks for the feedback, I'll try to get it done asap.

wagtail_localize/machine_translators/deepl.py
```diff
@@ -17,8 +27,9 @@ class DeepLTranslator(BaseMachineTranslator):
     display_name = "DeepL"
 
     def translate(self, source_locale, target_locale, strings):
+        api_endpoint = self.options.get("API_ENDPOINT", "https://api.deepl.com/v2/translate")
```
zerolab, line 30: nitpick: missed the first time around, but we do prefix our settings with `WAGTAILLOCALIZE_` so this should be consistent. i.e. `WAGTAILLOCALIZE_DEEPL_API_ENDPOINT`

vladox:
That doesn't seem to apply for AUTH_KEY and seems too verbose for an
option. Since it can also be used in the future for other endpoints I'd
leave it as it is.

zerolab:
> That doesn't seem to apply for AUTH_KEY and seems too verbose for an
option. Since it can also be used in the future for other endpoints I'd
leave it as it is.

@vladox sorry for the noise, I realised the options already came via `WAGTAILLOCALIZE_MACHINE_TRANSLATOR` after I made the comment, so removed it

wagtail_localize/tests/test_edit_translation.py
```diff
@@ -3212,6 +3212,37 @@ def test_cant_machine_translate_without_snippet_perms(self):
...
+        WAGTAILLOCALIZE_MACHINE_TRANSLATOR={
+            "CLASS": "wagtail_localize.machine_translators.deepl.DeepLTranslator",
+            "OPTIONS": {
+                "AUTH_KEY": "00000000-aaaa-bbbb-1111-cccccccccccc:fx",
+                "API_ENDPOINT": "https://api-free.deepl.com/v2/translate",
+            },
+        }
+    )
+    def test_deepl_machine_translate_page(self):
+        response = self.client.post(
```
zerolab, line 3225: Correct me if I'm wrong, but this will actually make and API request to the DEEPL endpoint. We should avoid this and instead mock the response

zerolab:
#604 provides a simpler approach to this, so may use that instead

zerolab:
Went with #604. Thank you very much for this @vladox

zerolab:
@vladox #604 conditionally loads the endpoint based on the key ([deepl library reference](https://github.com/DeepLcom/deepl-python/blob/dc44527b5416528779cd958d71f4b0ef5dbbb03e/deepl/util.py#L37-L40))

vladox:
@zerolab there's also a part of the PR that handles exceptions from the API and shows them correctly formatted to the end user, see `class ApiEndpointError(Exception)`.
I can do a separate PR only for that part since I think it's necessary to make the integration more user friendly.

zerolab:
@vladod you're right, let's do it. Just make sure you mock the request/response

PR #506, opened 2022-01-18
translation: in German in the wagtail admin buttons spaces are missing.

vladox:
e.g.

image


zerolab:
Hey @vladox , thank you for this. The translations are managed in Transifex, in the wagtail project. We fetch new translations before each release. Can you please contribute your changes there? Otherwise they will get lost.

vladox:
@zerolab thanks for the hint, it looks like somebody already did it 4 days ago: https://www.transifex.com/torchbox/wagtail/translate/#de/wagtaillocalize_js/308007195?q=text%3A'Published+on'
https://www.transifex.com/user/profile/th3hamm0r/
This can be closed I guess.

PR #507, opened 2022-01-18
Reset the has error flag on po import

### LINKED ISSUE #425: Translation errors can’t be fixed using Pontoon
STR:
- Send a translation using Pontoon with a translation error (e.g. going over the character limit)
- Send a shortened version of the translation using Pontoon

Current behavior:
Capture d’écran 2021-06-17 à 19 47 59
The translation is updated, but the error remains, and the character count still reflects the old translation’s length. The translation won’t get published until a Wagtail admin manually edits then save the translation.

Expected:
The error disappears as soon as a valid translation comes from Pontoon and is immediately published

zerolab:
Fixes #425


kaedroho (approved):
This looks good to me, thanks!

Issue #497, opened 2021-12-17
Unauthorized PUT request when saving translated string


davidsevcik:
Hi all, 
I'm getting 401 Unauthorized server response when trying to save an individual string in Wagtail translation UI

![Screenshot 2021-12-17 at 12 04 05](https://user-images.githubusercontent.com/116192/146536376-c9ea4539-10c3-4843-b692-8346eb6e4cd1.png)

My dependencies:
```
Django==3.2.10
wagtail==2.15.1
wagtail-localize==1.0.1
```

Does anybody else experience this?

zerolab:
The string translation url is an admin URL ([ref](https://github.com/wagtail/wagtail-localize/blob/e782c78fcee5cd85fde5ca8bb3b3c48e857cfda8/wagtail_localize/wagtail_hooks.py#L63))

The view itself should only return a 403 is the requesting user https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/views/edit_translation.py#L986-L1040


Is your site (Settings > Sites) properly configured? Do you have some kind of aggressive ad blocker or similar browser extension? Does this work in a different browser at all?

Personally I have not had this issue, be it locally or on a staging or production site

hennabanana:
+1

hennabanana:
Machine Translation works. i compared both requests, in the the `PUT | http://localhost:8000/admin/localize/translate/142/strings/981/edit/` the csrfmiddlewaretoken is missing. But in the Machine Translation PUT it is there.



![Screenshot 2021-12-30 at 10 11 52](https://user-images.githubusercontent.com/25147943/147738123-a465ee21-a483-47f5-a6eb-d3e2925f7da6.png)
![Screenshot 2021-12-30 at 10 11 31](https://user-images.githubusercontent.com/25147943/147738131-651d22fd-e1b2-46a9-9358-296965095afe.png)

hennabanana:
it has to do with the django rest framework, when i comment out the permissions in the settings it works.

`REST_FRAMEWORK = {
    # 'DEFAULT_AUTHENTICATION_CLASSES': (
    #     'rest_framework.authentication.TokenAuthentication',
    #     'rest_framework.authentication.BasicAuthentication',
    # ),
}`

zerolab:
@hennabanana ah! could you maybe give #500 a try?

hennabanana:
@zerolab  https://github.com/wagtail/wagtail-localize/pull/512

[...]

PR #512, opened 2022-01-27
Add Django-Rest-Framework Settings to Readme

hennabanana:
Its enough to set these settings :)

Issue: [#497](https://github.com/wagtail/wagtail-localize/issues/497)

Pull: [500](https://github.com/wagtail/wagtail-localize/pull/500)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,18 @@ or, you can run them for a specific environment `tox -e python3.8-django3.2-wagt
 
 To run the test app interactively, use `tox -e interactive`, visit `http://127.0.0.1:8020/admin/` and log in with `admin`/`changeme`.
 
+### Django Rest Framework
+
+If you use Django Rest Framework, be sure that you add the `SessionAuthentication` in your `settings.py`
+
+```python
+REST_FRAMEWORK = {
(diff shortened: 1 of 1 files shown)


zerolab:
Hey @hennabanana, 
I appreciate this PR, however we should not force `rest_framework.authentication.SessionAuthentication` as the authentication backend on all projects as such. The proper fix is to go with the suggestion from #499 which #500 is trying to fix.

Can you give the #500 a try?

zerolab:
Ah, I see it doesn't work for you 🤔
Will have a further look into this tomorrow. Leaving the PR open for now

zerolab:
As per https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme, we want to use `@authentication_classes([SessionAuthentication])` on https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/views/edit_translation.py#L987 rather than require it globally.

Will address as part of the work on #500. Closing this PR on that basis. Thank you @hennabanana

zerolab:
Should be fixed by #513

PR #513, opened 2022-02-01
Fix `edit_string_translation` and `edit_override` views when DRF is configured with explicit permission/authentication classes

### LINKED ISSUE #497: Unauthorized PUT request when saving translated string

### LINKED ISSUE #499: edit_string_translation view should not use REST_FRAMEWORK default permissions and authentication classes

zerolab:
Continuation of #500 with test fixes.
Fixes #499, fixes #497

### DIFF
--- a/wagtail_localize/views/edit_translation.py
+++ b/wagtail_localize/views/edit_translation.py
@@ -19,7 +19,12 @@ from django.utils.translation import gettext as _
 from django.views.decorators.http import require_POST
 from modelcluster.fields import ParentalKey
 from rest_framework import serializers, status
-from rest_framework.decorators import api_view, permission_classes
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.decorators import (
+    api_view,
+    authentication_classes,
+    permission_classes,
+)
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from wagtail.admin import messages
@@ -986,6 +991,7 @@ def restart_translation(request, translation, instance):
 
 @api_view(["PUT", "DELETE"])
 @permission_classes([IsAuthenticated])
+@authentication_classes([SessionAuthentication])
 def edit_string_translation(request, translation_id, string_segment_id):
     translation = get_object_or_404(Translation, id=translation_id)
     string_segment = get_object_or_404(StringSegment, id=string_segment_id)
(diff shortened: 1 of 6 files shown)


wagtail_localize/views/edit_translation.py
```diff
@@ -984,6 +990,8 @@ def restart_translation(request, translation, instance):
 
 
 @api_view(["PUT", "DELETE"])
+@permission_classes([IsAuthenticated])
+@authentication_classes([SessionAuthentication])
```
chris48s, line 994: The one thing I'll quickly throw into the mix here is that IIUC this overrides rather than appends, so `edit_string_translation` and `edit_override` endpoints kind of become "internal only" API endpoints. i.e: if I wanted to explicitly write a third party app that interacts with those endpoints using token auth or JWTs or whatever, I now can't do it. If we wanted to enable that use-case, we could do something like

```py
@authentication_classes([SessionAuthentication] + settings.DEFAULT_AUTHENTICATION_CLASSES)
```

I don't quite have the context to say whether that is something we would want or not, so I'll defer to your judgement but just thought I'd flag it.

chris48s: cool. Sounds sensible then :+1:

chris48s (approved):

bmihelac:
@zerolab thanks!

zerolab:
My pleasure. Aiming for a new release sometime next week.

PR #518, opened 2022-02-08
Redirect to newly translate page's edit view when submitting a translation for a single locale

mixxorz:
When an editor decides to submit a page for translation and only chooses one locale, they are redirected to the new translated page's edit view.

### DIFF
--- a/wagtail_localize/tests/test_submit_translations.py
+++ b/wagtail_localize/tests/test_submit_translations.py
@@ -78,7 +78,7 @@ class TestTranslatePageListingButton(TestCase, WagtailTestUtils):
         self.assertContains(
             response,
             (
-                f'<a href="/admin/localize/submit/page/{self.en_blog_index.id}/?next=%2Fadmin%2Fpages%2F{self.en_homepage.id}%2F" '
+                f'<a href="/admin/localize/submit/page/{self.en_blog_index.id}/" '
                 'aria-label="" class="u-link is-live ">'
                 "\n                    Translate this page\n                </a>"
             ),
@@ -624,7 +624,7 @@ class TestTranslateSnippetListingButton(TestCase, WagtailTestUtils):
         self.assertContains(
             response,
             (
-                f'href="/admin/localize/submit/snippet/wagtail_localize_test/testsnippet/{self.en_snippet.id}/?next=%2Fadmin%2Fsnippets%2Fwagtail_localize_test%2Ftestsnippet%2F"{extra}>Translate</a>'
+                f'href="/admin/localize/submit/snippet/wagtail_localize_test/testsnippet/{self.en_snippet.id}/"{extra}>Translate</a>'
             ),
         )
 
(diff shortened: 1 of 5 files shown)


wagtail_localize/views/submit_translations.py
```diff
@@ -222,7 +222,16 @@ def _walk(current_page):
         # TODO: Button that links to page in translations report when we have it
         messages.success(self.request, self.get_success_message(locales))
 
-        return redirect(self.get_success_url() or self.get_default_success_url())
+        single_translated_object = None
+        if len(form.cleaned_data["locales"]) == 1:
```
zerolab, line 226: nitpick: we have a `if len(form.cleaned_data["locales"]) == 1:` check just above which preps the locale message, so perhaps merge the two?

wagtail_localize/tests/test_submit_translations.py
```diff
@@ -807,6 +804,14 @@ def test_post_submit_snippet_translation(self):
...
         self.assertEqual(translated_snippet.field, "Test snippet")
 
+        self.assertRedirects(
+            response,
+            reverse(
+                "wagtailsnippets:edit",
+                args=["wagtail_localize_test", "testsnippet", translated_snippet.id],
+            ),
+        )
+
```
zerolab, line 814: suggestion: add a test to check that submitting a translation to 2+ locales takes you to the previous locations

zerolab (changes requested):
Nice one @mixxorz.
Got a couple of minor changes then this is good to go

wagtail_localize/tests/test_submit_translations.py
mixxorz: These two tests already cover that scenario so I don't think we need to add a new test for this.

* `test_post_submit_page_translation_into_multiple_locales`
* `test_post_submit_snippet_translation_into_multiple_locales`

wagtail_localize/views/submit_translations.py
mixxorz: Good call. I've merged them now!

wagtail_localize/views/submit_translations.py
```diff
@@ -222,7 +226,10 @@ def _walk(current_page):
         # TODO: Button that links to page in translations report when we have it
         messages.success(self.request, self.get_success_message(locales))
 
-        return redirect(self.get_success_url() or self.get_default_success_url())
+        return redirect(
+            self.get_success_url()
+            or self.get_default_success_url(single_translated_object)
```
zerolab, line 231: I've just tested this locally.
This bit is potentially confusing as the only way you can get to the translation edit page as it stands is if you tree syncing enabled and you click "translate this page" on the alias edit view

When you use the "more" actions from a page listing, `self.get_success_url()` is non-empty because there's a `next` query param with a value going back to the listing page (e.g. `admin/localize/submit/page/3/?next=%2Fadmin%2Fpages%2F60%2F`)

so perhaps move the single_translated_object to `get_success_url` ?

zerolab: alternatively, remove the next_url handling from https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/wagtail_hooks.py#L153-L154 🤔

mixxorz: I went with your second suggestion. I think it makes sense that we remove the "next" param from the "Translate this page" button as the view doesn't need it to know where to redirect the editor next.

This also means that we can keep the "next URL routing" feature if we needed it at some point.

PR #523, opened 2022-02-16
Handles some fringe cases after DB migrations

vladox:
Some DB migrations may leave fields with NULL values that need proper handling.

### DIFF
--- a/wagtail_localize/segments/types.py
+++ b/wagtail_localize/segments/types.py
@@ -80,6 +80,9 @@ class StringSegmentValue(BaseValue):
             attrs (dict, optional): A dict of HTML attributes that were stripped out of the string.
             order (int, optional): The index that this segment appears on a page.
         """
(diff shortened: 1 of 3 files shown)


zerolab:
@vladox thanks for this. Looks like `StringSegmentValue` should only accept strings, based on the failing test (`wagtail_localize.segments.tests.test_segmentvalue.TestStringSegmentValue.test_raises_typeerror_if_string_is_none`). imho the special casing should happen at the extract/ingest point

zerolab (changes requested):
@vladox thanks for this. Looks like StringSegmentValue should only accept strings, based on the failing test (wagtail_localize.segments.tests.test_segmentvalue.TestStringSegmentValue.test_raises_typeerror_if_string_is_none). imho the special casing should happen at the extract/ingest point

zerolab:
Closing in favour of #621. Thank you @vladox

PR #525, opened 2022-02-22
Support `ListBlock`s nested in `StructBlock`s

zerolab:
A continuation of #510, this adds support for `ListBlock` in `StrucBlock` and expands the test case for `ListBlock` in nested `StreamBlock`


kaedroho (approved):
This looks good to me, thanks!

Do you think we would need to to a similar thing for list block in nested stream blocks too?

PR #526, opened 2022-02-22
Prep v1.1rc1 release

zerolab:


th3hamm0r:
@zerolab The i18n .mo files haven't been compiled yet. I think the compilation task has been missed for the 1.0.1 release (and maybe earlier) too, all .mo files are older than their related .po files.

I stumbled over this because with the currently released 1.0.1 version some german some translations are missing in the UI and I always thought they have been added after 1.0.1 has been released, but actually they have been added to the PO files before.

zerolab:
Thank you for the heads up @th3hamm0r. Will revisit the release prep steps as something is out of order. Will make another release on Friday at the latest

Issue #402, opened 2021-06-03
Deleting translated pages/snippets/etc does not delete their data


Pomax:
It looks like wagtail-localize does not clean up the data associated with translated pages/snippets when deleting, allowing for the following situation:

- create a translation of a page
- localise the content
- publish the localized page
- delete the localized page
- create a new translation of original page for the same locale as before
- all fields are suddenly already translated, using the previous translations

Instead, deleting a localized page should also delete all the associated localized fields, so that a new translation is a proper, new translation.

th3hamm0r:
This issue hit us together with the problem described in #524:
We had a mistyped translation of a subtitle, but it was decided to remove the subtitle completely from the (source) page. Now the source page had no subtitle, but the synced translated page still showed the mistyped string, which was only displayed in the frontend, but not editable in the CMS. The first attempt to fix it, was to remove the translated page and re-create it, but this did not work as described by @Pomax above, since the translations seem to be re-used for the recreated translation...

I think too, that deleting a translation should remove those localized fields, so re-creating a translation should not display previous translations.

PR #532, opened 2022-02-25
Delete the related Translation object when converting a page back to alias

zerolab:
Follow up to #515

Doing a bit of tidy up and is in line with #402 (although this does not fix that particular issue)

### DIFF
--- a/wagtail_localize/views/convert.py
+++ b/wagtail_localize/views/convert.py
@@ -70,7 +70,9 @@ def convert_to_alias(request, page_id):
                 },
             )
 
-            # Now clean up the old translation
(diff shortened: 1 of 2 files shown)


wagtail_localize/views/convert.py
```diff
@@ -70,6 +70,15 @@ def convert_to_alias(request, page_id):
                 },
             )
 
+            # Now clean up the old translation
```
kaedroho, line 73: Probably worth noting here why we delete instead of disable?

zerolab: good shout. done

PR #545, opened 2022-03-18
extract.py: fix list block for empty lists

mb03:
In case there is a listblock with no elements, by accident or design, the current code would raise an error.
This fix solves the case of empty list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 /node_modules
 /test-static
 /test-media
+.idea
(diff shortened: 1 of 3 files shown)


zerolab:
Nice catch, @mb03
Can you please [lint](https://github.com/wagtail/wagtail-localize#pre-commit) the code and add a test for this too?

You can see #510 or #525 for examples of tests for it

mb03:
Linted and added a test

zerolab:
Merged, thank you

Issue #287, opened 2020-12-21
Version 0.9.4 doesn't give possibility to translate pagechooserblock content


ciloeweb:
Hello, I implemented a page which contain a pagechooserblock inside Streamfield. I have two locales English and Spanish, I started using wagtail-localized version 0.9.2 and the library related each page with the corresponding locale page and it give me the possibility to edit as you see below:

English Page:
![image](https://user-images.githubusercontent.com/6557784/102832094-14acad80-43b3-11eb-9f30-6d237f0bbbbf.png)

Spanish Page
![image](https://user-images.githubusercontent.com/6557784/102832129-27bf7d80-43b3-11eb-96fa-fdb5895acf61.png)

But when I updated to 0.9.4 version, the library just adds the page id and use the English version without the possibly to edit, it is using the English version.

![image](https://user-images.githubusercontent.com/6557784/102832144-3148e580-43b3-11eb-9db7-4ae55ea80026.png)

I wonder if this is a redesign of the library or it is a bug? Because now I can’t get the content translated.

zerolab:
This was fixed in #480 and available in version 1.0

olifante:
The bug still seems to be there when you subclass StreamBlock and use the PageChooserBlock in your subclass, instead of using it directly in your StreamField. I'm using `wagtail-localize` 1.1 on Wagtail 2.16.1.

Here’s a screenshot showing the differing behaviour between these two cases. The two tasks at the bottom which are incorrectly displaying a page ID of 3 are using a StreamBlock subclass, unlike the otherwise very similar page choosers above it.

![Screenshot 2022-03-22 at 11 04 44](https://user-images.githubusercontent.com/5111/159468019-54a251da-e222-4ea4-a7f9-5de41e752b54.png)

zerolab:
@olifante I've opened #546 that fixes ^.
Made a build at https://zerolab.org/_dumps/wagtail_localize-1.1.1a1-py3-none-any.whl
You can install it via `pip install https://zerolab.org/_dumps/wagtail_localize-1.1.1a1-py3-none-any.whl\#egg\=wagtail_localize` to test locally

olifante:
That build fixes the problem with my code example (PageChooserBlock in a StreamBlock), but I'm still having problems with other pages. One of them happens when we're using a PageChooserBlock inside a StreamBlock inside a StructBlock. I'm going to try to create a minimal example based on our code that highlights the remaining error.

zerolab:
[...]

PR #480, opened 2021-11-01
Allow overriding chooser blocks defined in `StructBlock`

zerolab:
Fixes #448

To-Do:

- [x] tests
- [x] nested `StructBlock`s

PR #546, opened 2022-03-23
Handle choosers in nested `StreamBlock`s

zerolab:
This fixes https://github.com/wagtail/wagtail-localize/issues/287#issuecomment-1075032672 as a follow up to #480

To-Do (probably as a follow up PR as it feels like a deep rabbit hole):

- [x] chooser in `ListBlock` in nested `StreamBlock`
- [x] check chooser in `StructBlock` in `ListBlock` in `StreamBlock`


wagtail_localize/models.py
```diff
@@ -1511,7 +1511,7 @@ def get_field_path_from_stream_block(stream_value, path_components):
                         ]
 
                         if isinstance(block_def, blocks.StructBlock):
-                            return [block.block_type, path_components[1]]
+                            return [block.block_type] + path_components[1:]
```
zerolab, line 1514: This change ensures we can process a chooser (e.g. image) in a StructBlock in a StructBlock in a StreamBlock in a StreamBlock


e.g.

```python
class ImageBlock(StructBlock):
    """
    Custom `StructBlock` for utilizing images with associated caption and
    attribution data
    """
    image = ImageChooserBlock(required=True)
    caption = CharBlock(required=False)


class Card(StructBlock):
    image_struct_block = ImageBlock()
    page = PageChooserBlock()
    text = CharBlock(required=False)


class PageInStreamBlock(StreamBlock):
    text = CharBlock(required=False)
    page_link = PageChooserBlock()
    card = Card()


class StoryBlock(StreamBlock)

    pages_in_streamblock = PageInStreamBlock()
    nested_stream = StreamBlock([
        ('pages_in_streamblock', PageInStreamBlock()),
    ])
```

wagtail_localize/views/edit_translation.py
```diff
@@ -414,11 +414,17 @@ def widget_from_block(block, content_components=None):
...
         if isinstance(block_type, blocks.StructBlock):
             block_field_name = field_path_components[2]
             block_field = block_type.child_blocks[block_field_name].label
-
+        elif isinstance(block_type, blocks.StreamBlock):
+            block_field_name = field_path_components[2]
+            block_field = block_type.child_blocks[block_field_name].label
+            # use the field path components instead of content path as we're drilling down
+            # a set of nested blocks
+            content_components = field_path_components[2:]
```
zerolab, line 427: this then makes use of the change in models.py

wagtail_localize/models.py
kaedroho: It looks like this would work for a structblock in a structblock because the field and content paths are the same for struct blocks.

If there's a streamblock within a structblock, would the content paths be converted to the correct field paths?

wagtail_localize/views/edit_translation.py
kaedroho: A little worried about mixing field/content paths since they will have different values when working with nested streamblocks.

 - Content path is a path within an instance, so stream blocks are represented with their ID
 - Field path is a path on the data model, so stream blocks are represented with their type name (this was introduced to allow us to look up the field a segment was inserted into without having to retrieve the original content from the database)

So if you pass a field path into a function that expects a content path, it'll only work if there aren't any streamblocks within that path. So I think here it could break if there's a nested stream block?

zerolab: Note: tidied up in 1e284bb (during our pair programming session)

wagtail_localize/models.py
zerolab: Note: tidied up in 1e284bb (during our pair programming session)

also added a test

PR #550, opened 2022-04-01
Translatable ModelAdmin

### LINKED ISSUE #291: ModelAdmin doesn't have Locale
I can create snippets items for different languages - they have option - Locale (I can choose language in right corner and furthermore I can translate button)

Screenshot from 2020-12-29 12-05-30

But if I want to create snippet item through **_wagtail.contrib.modeladmin.options.ModelAdmin** - I can't use Locale and translate features (I don't have language chooser in right corner and furthermore I don't have translate button)
Screenshot from 2020-12-29 12-06-11

I really need list_display, search, filter - so I have to use ModelAdmin
But at the same time - I need the possibilities to create multi-language snippets 

maybe I do something wrong and we can create locale ModelAdmin ???

dinoperovic:
This PR adds support for translating pages and models registered using Wagtail's `modeladmin`. 
It mostly re-uses the snippets approach and exposes the `TranslatableModelAdmin` class.

Translatable ModelAdmin has been a requirement in my personal projects and I have been using a similar setup that works pretty well.

Fixes #291.

A less intrusive approach that I was considering is to release a separate `wagtail-localize-modeladmin` package. In 
this case the snippet views can be "patched" by placing the app above `wagtail.snippets` and handling redirects to `modeladmin` instances when applicable.

Curious to hear your thoughts and plans for supporting ModelAdmin localization.


dinoperovic:
@zerolab Yeah I've missed that one. It looks like this could replace some parts of view logic. The views would still need to be overridden to display translation pages etc. Also, to support the previous Wagtail versions. It would be a good idea to align with the approach in the https://github.com/wagtail/wagtail/pull/7857 once it's accepted. 

I can take a closer look next week.

zerolab:
cc @kaedroho (in the context of modeladmin/snippets thinking)

wagtail_localize/modeladmin/helpers.py
```diff
@@ -0,0 +1,78 @@
...
+        if source is not None and source.translations.filter(enabled=True).exists():
+            url = reverse("wagtail_localize:update_translations", args=[source.id])
+            if next_url is not None:
+                url += "?" + urlencode({"next": next_url})
+
+            yield {
+                "url": url,
+                "label": _("Sync translated %s") % obj._meta.verbose_name_plural,
+                "classname": classname,
+                "title": _("Sync translated %s") % obj._meta.verbose_name_plural,
```
zerolab, line 77: Suggestion: use named placeholders. Something like `_("Sync translated %(model_name)s") % {"model_name": obj._meta.verbose_name_plural}`

It is a pattern we use in Wagtail core and in localize

wagtail_localize/modeladmin/options.py
```diff
@@ -0,0 +1,62 @@
+from django.core.exceptions import ImproperlyConfigured
+from wagtail.contrib.modeladmin.options import (
+    ModelAdmin,
+    ModelAdminGroup,
+    modeladmin_register,
+)
```
zerolab, line 6: qustion: why are these imported and declared in `__all__`?
I see the test wagtail_hooks imports it from here, rather than `wagtail.contrib.modeladmin`

wagtail_localize/modeladmin/views.py
```diff
@@ -0,0 +1,164 @@
...
+        context["locale"] = self.locale
+        context["translations"] = []
+        context["wagtail_version"] = get_main_version()
+        return context
+
+
+class TranslatableIndexView(TranslatableViewMixin, IndexView):
+    def get_filters(self, request):
+        filters = super().get_filters(request)
+        filters[2]["locale"] = self.locale.id
```
zerolab, line 59: nitpick: this warrants at least a comment as to why `filters[2]`. Had to double check myself. `get_filters()` returns `(filter_specs, bool(filter_specs), lookup_params, use_distinct)`

wagtail_localize/modeladmin/views.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class TranslatableEditView(TranslatableViewMixin, EditView):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.translation = Translation.objects.filter(
+            source__object_id=self.instance.translation_key,
+            target_locale_id=self.instance.locale_id,
+        ).first()
```
zerolab, line 102: nitpick: this does need a comment as to why `.first()` for future reference

wagtail_localize/views/edit_translation.py
```diff
@@ -872,13 +906,24 @@ def get_translation_progress(segment, locale):
...
+                                    ],
+                                )
+                                if translated_instance._meta.model
+                                in get_snippet_models()
+                                else reverse(
+                                    "{app_label}_{model_name}_modeladmin_edit".format(
+                                        app_label=translated_instance._meta.app_label,
+                                        model_name=translated_instance._meta.model_name,
+                                    ),
+                                    args=[translated_instance.id],
```
zerolab, line 925: I feel this block is convoluted enough to warrant extracting to own method

wagtail_localize/views/submit_translations.py
```diff
@@ -317,3 +318,38 @@ def get_success_message(self, locales):
             object=str(self.object),
             locales=locales,
         )
+
+
+class SubmitModelAdminTranslationView(SubmitTranslationView):
```
zerolab, line 323: imho this should live in `.modeladmin.views` since `modeladmin` is an optional sub-app

wagtail_localize/models.py
```diff
@@ -126,6 +127,15 @@ def get_edit_url(instance):
             ],
         )
 
+    else:
+        return reverse(
```
zerolab, line 131: note: this implies one does use modeladmin which won't necessarily be the case.

My take on this is: `wagtail_localize.modeladmin` should be optional. If not added to `INSTALLED_APPS` then this the `SubmitModelAdminTranslationView` vuew and `submit_modeladmin_translation` path declaration should not register and come into play at all

zerolab (changes requested):
Hey @dinoperovic,

This is a lot of excellent work, thank you. 
Left a number of questions and comments.

Additionally, this could really use some documentation

wagtail_localize/modeladmin/options.py
dinoperovic: This was just a re-export for convenience but is unnecessary. I've removed it.

wagtail_localize/modeladmin/views.py
dinoperovic: Added the extra comment.

wagtail_localize/modeladmin/views.py
dinoperovic: Added the extra comment.

wagtail_localize/views/edit_translation.py
dinoperovic: Extracted into a new `get_edit_url` func, also added the `get_submit_translation_url` function as well.

wagtail_localize/views/submit_translations.py
dinoperovic: Refactored modeladmin as an optional sub-app and moved the views and tests to modeladmin module.

dinoperovic:
Hey @zerolab,

I've made the requested changes, still missing the documentation which I will look to add.

wagtail_localize/views/edit_translation.py
zerolab: ah, that reads so so much better! thank you

wagtail_localize/templates/wagtail_localize/modeladmin/includes/button.html
```diff
@@ -0,0 +1 @@
+<a{% if button.url %} href="{{ button.url }}{% if locale %}?locale={{ locale.language_code }}{% endif %}"{% endif %} class="{{ button.classname }}" title="{{ button.title }}"{% if button.target %} target="{{ button.target }}"{% endif %}{% if button.rel %} rel="{{ button.rel }}"{% endif %}>{{ button.label }}</a>
```
kaedroho, line 1: Could we move these templates into the ``wagtail_localize.modeladmin`` app?

kaedroho (commented):
This is excellent work, thanks! Just one minor thing I wanted to question

wagtail_localize/templates/wagtail_localize/modeladmin/includes/button.html
dinoperovic: Yes, that makes sense, moved it!

kaedroho (approved):

zerolab:
Merged. Thank you @dinoperovic. Will make a new release sometime towards the end of the week

PR #559, opened 2022-04-28
Support further `ListBlock` nesting

zerolab:
`ListBlock` in `StructBlock` in `StructBlock` in `ListBlock` (and probably a few further levels down the rabbit hole`

Test build: https://zerolab.org/_dumps/wagtail_localize-1.1.2rc1-py3-none-any.whl


wagtail_localize/segments/extract.py
```diff
@@ -162,10 +162,21 @@ def handle_list_block(self, list_block, raw_value=None):
...
             if has_block_format:
-                for block in list_block.bound_blocks:
+                for index, block in enumerate(list_block.bound_blocks):
+                    # pass on the relevant sub-block raw value, should it be a ListBlock
+                    try:
+                        if isinstance(raw_value, list):
+                            block_raw_value = raw_value[index]
+                        elif isinstance(raw_value, dict):
+                            block_raw_value = raw_value["value"][index]
+                    except (IndexError, KeyError):
```
chris48s, line 172: Are `list` and `dict` the only two types `raw_value` could possibly have here? If not, there's a risk `block_raw_value` could be undeclared when we try to reference it later.

We could make this either:

```py
block_raw_value = None
try:
    if isinstance(raw_value, list):
        block_raw_value = raw_value[index]
    elif isinstance(raw_value, dict):
        block_raw_value = raw_value["value"][index]
except (IndexError, KeyError):
    block_raw_value = None
```

or

```py
try:
    if isinstance(raw_value, list):
        block_raw_value = raw_value[index]
    elif isinstance(raw_value, dict):
        block_raw_value = raw_value["value"][index]
    else:
        block_raw_value = None
except (IndexError, KeyError):
    block_raw_value = None
```

to guard against that.

chris48s (commented):
I'm assuming the netlify build failures are unrelated to these functional changes?

wagtail_localize/segments/extract.py
zerolab: went with second option.
raw_value would always be a list or a dict at this point. it can be None, but we exit early. Nonetheless it doesn't hurt to be more protective.

The Netlify errors are unrelated -- we have them in last few PRs. Will have a look at them when I have more time

chris48s (approved):

PR #560, opened 2022-04-29
Fix header styling for Wagtail 3.0

kaedroho:
This PR adds back CSS that was removed in Wagtail 3.0 to make the header look as it was before.

I think it'll be simpler to use the old header styling until we drop support for Wagtail 2. What do you think?

**Wagtail 2.16**

image

**Wagtail 3.0 (RC3)**

image


zerolab (approved):
I think this is a good intermediary step. Makes the edit interface usable in 3.0, at least.

PR #569, opened 2022-05-06
3.0 official support and catch with wagtailmain

zerolab:


zerolab:
@kaedroho this should sort out the nightly failures against Wagtail main

PR #572, opened 2022-05-20
Fix further Wagtail `main` failures and tidy up GH Action/Makefile

zerolab:
This PR:

* adds a callable for the page revision model (with Wagtail 4.0, that will be `wagtailcore.Revision` ref: https://github.com/wagtail/wagtail/pull/8441) -- should fix the nightly test failures
* updates the test GitHub Action to:
  * cancel previous in-progress jobs for the same branch/PR
  * cache tox
  * use newer action versions from GitHub
* adds a few more commands to the Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,47 @@
-npm_deps:
+# ----------------------------------------------------------------------------
+# Self-Documented Makefile
+# ref: http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
+# ----------------------------------------------------------------------------
+.PHONY: help
+.DEFAULT_GOAL := help
+
+help:											## ⁉️  - Display help comments for each make command
+	@grep -E '^[0-9a-zA-Z_-]+:.*? .*$$'  \
+		$(MAKEFILE_LIST)  \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,18 +1,47 @@
...
+
+clean:	## 🗑️  - Remove __pycache__ and test artifacts
+	@echo "🗑️ - Removing __pycache__ and test artifacts"
+	find . -name ".tox" -prune -o -type d -name  "__pycache__" -exec rm -r {} +
+
+package-setup:
+	@echo "📦 - Packaging for PyPI"
+	python3 setup.py clean --all sdist bdist_wheel
+
+package: clean package-setup  ## 📦 - Package for PyPI
```
chris48s, line 43: Question: should we also run `translations` as part of the package build?

chris48s (commented):
One question but aside from that LGTM. If the answer is 'no', just merge it :+1:

Makefile
zerolab: the process is:
- make messages, fetch translations, compilemessages -> commit
- update version, changelog -> commit
- tag and package

so can't quite include it

PR #578, opened 2022-05-24
updates filter to get translation source

### LINKED ISSUE #547: MultipleObjectsReturned - multiple TranslationSource for one translation_key
Hello,
I somehow managed to get an inconsistency in my database. When editing a Page I now get the error  `MultipleObjectsReturned` for `TranslationSource` which is called from `wagtail_localize/wagtail_hooks.py` from `ConvertToAliasPageActionMenuItem._is_shown`:

```python
TranslationSource.objects.get(
                        object_id=page.translation_key,
                        specific_content_type=page.content_type_id,
                    ).locale_id
```

What I'm wondering is, whether this is actually valid to have multiple `TranslationSource` for the same `translation_key`? If not it might make sense to add an additional unique index on the database level? right now there is only a unique constraint for `("object", "locale"),` .. maybe it would make more sense to have `("object", "specific_content_type")` as unique constraint, so this error can't occur?

(I think i created the problem by translating a child page from german to english, while the parent page was still a page alias, and not yet translated.. and then somehow managed to accidentally translate this english alias page to english.. which is another problem I guess.. when there is e.g. a german page with an english alias.. the english list would still show the "Translate this page" option in the "More" menu.. which creates a english -> english `Translation` object..)

sheralim012:
Fixes #547 

## steps to reproduce the issue
- create locales L1, L2, L3
- create a page P1 in locale L1
- translate page P1 to locale L2 (let's say the newly translated page is P1')
- translate page P1' to locale L3 (let's say the newly translated page is P1'')
- you will be redirected to edit page with error `MultipleObjectsReturned - get() returned more than one TranslationSource -- it returned 2!`

## observations
- above-mentioned scenario creates 2 translation sources with same `object_id`, and `spectific_content_type_id` but from different locales
- "convert to alias" buttons' logic (finding the source of translation) is dependent on these 2 fields but these 2 fields alone can't uniquely identify the translation source (hence the error - multiple objects returned)

## solution
the combination of `object_id`, `spectific_content_type_id`, and `translations__target_locale_id` can uniquely identify the translation source

Let me know if I missed something.

### DIFF
--- a/wagtail_localize/tests/test_convert_to_alias.py
+++ b/wagtail_localize/tests/test_convert_to_alias.py
@@ -99,6 +99,20 @@ class ConvertToAliasTest(ConvertToAliasTestData, TestCase):
             reverse("wagtail_localize:convert_to_alias", args=[self.fr_page.id]),
         )
 
+    def test_page_action_available_for_translated_page_from_translated_page(self):
+        de_locale = Locale.objects.create(language_code="de")
+        self.client.post(
+            reverse(
+                "wagtail_localize:submit_page_translation",
+                args=[self.fr_page.id],
+            ),
+            {"locales": [de_locale.id]},
+        )
+
+        de_page = self.fr_page.get_translation(de_locale)
+        de_page.refresh_from_db()
+        self.assertTrue(self._page_action_is_shown(de_page))
+
 
 class ConvertToAliasViewTest(ConvertToAliasTestData, TestCase):
     def setUp(self):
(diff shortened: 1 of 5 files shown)


zerolab:
Hey @sheralim012, thank you for this!
Any chance you could add some tests?

sheralim012:
@zerolab I have added the test cases.
Could you review it now?

zerolab (approved):

PR #580, opened 2022-05-27
Fix duplicate locale defintion in TranslatableCreateView template

### LINKED ISSUE #579: TranslatableCreateView gets Error 404 after submit
When I use the TranslateableModelView, I can edit and also delete entries without any problems.
But when I try to create new entries, it always fails with a 404 error after submitting the form.

The reason is that the "?locale=en" suffix is appended twice to the URL : 
`<form action="/admin/{app}/{model}/create/?locale=en?locale=en" method="POST" novalidate="">`

This causes Django to misinterpret the URL and stumble.

benmth:
This PR fixes #579, as it removes the duplicate definition of the "locale" parameter. 

The "?locale=" parameter is already attached to the URL in the base template ("modeladmin/create.html"):
`{% block form_action %}{{ view.create_url }}{% endblock %}{% if locale %}?locale={{ locale.language_code }}{% endif %}`

As a result, it is currently not necessary to override the "form_action" block.

Removing the code ensures that the CreateView works again, and does not fail because of a malformed URL, ending in a 404 error.

### DIFF
--- a/wagtail_localize/modeladmin/templates/wagtail_localize/modeladmin/translatable_create.html
+++ b/wagtail_localize/modeladmin/templates/wagtail_localize/modeladmin/translatable_create.html
@@ -1,7 +1,5 @@
 {% extends "modeladmin/create.html" %}
 
 {% block header %}
-  {% include "wagtailadmin/shared/header_with_locale_selector.html" with title=view.get_page_title subtitle=view.get_page_subtitle icon=view.header_icon tabbed=1 merged=1 %}
+    {% include "wagtailadmin/shared/header_with_locale_selector.html" with title=view.get_page_title subtitle=view.get_page_subtitle icon=view.header_icon tabbed=1 merged=1 %}
 {% endblock %}
-
-{% block form_action %}{{ view.create_url }}?locale={{ locale.language_code }}{% endblock %}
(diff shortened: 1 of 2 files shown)


zerolab:
Hey @benmth, thanks for this. Could you please add test for the fix?
This will ensure future changes don't undo the work

zerolab:
I tried to replicate the issue locally and cannot, so this most certainly needs test

benmth:
Hey @zerolab,

i am not very experienced with testing, especially when it comes to Django templates.

Due to little time at the moment, a test from me will therefore unfortunately take some time, because I have to get acquainted first to be able to deliver something useful.

But maybe this information will help you to reproduce the original problem:

It occurs in my existing project, as well as in a newly created (with the latest wagtail and wagtail-Localize) versions. The two locales are English and French. 

In my existing project the model in question is defined as a snippet at the same time, in my debug project I'm unsure right now. 

If I then define a TranslatableModelAdmin for the model in question, I can delete and edit entries, but not create them. The edit form appears, but a save fails because the form action contains 2x "?locale=xxx".

benmth:
again, sorry to keep you waiting.

I found the time to look at it today and was able to expand the tests accordingly.

In doing so, I then also extended the original change to respect Wagtail installations under version 3.0.

If there is still something to do from my side, please just let me know.

zerolab (approved):
Nice on @benmth!

zerolab:
version 1.2.1 is now out

PR #589, opened 2022-07-01
Use Flit for packaging

engineervix:
Starting with [PEP 621](https://peps.python.org/pep-0621/), the Python community selected `pyproject.toml` as a standard way of specifying project metadata. Given that the future of Python packaging is `pyproject.toml` (based on [PEP 518](https://peps.python.org/pep-0518/) and [PEP 621](https://peps.python.org/pep-0621/), transitioning Python projects to use `pyproject.toml` is inevitable. 

This PR introduces [Flit](https://flit.pypa.io/) as a packaging tool for the project. Flit prides itself in its provision of a simple way to create and upload pure Python packages and modules to PyPI. It focuses on [making the easy things easy](https://flit.readthedocs.io/en/latest/rationale.html) for packaging.

TODO:

- [x] Update README / docs with build instructions
- [x] Update the relevant `Makefile`targets to incorporate Flit


chris48s:
Nice. I'll have a proper look over this next week, but I'll flag a couple of quick things now. One is we need to get the netlify builds working:

https://github.com/wagtail/wagtail-localize/blob/main/netlify.toml

Tbh, it should probably work as-is if we upgrade pip:

```
pip install --upgrade pip
pip install -e.[testing,documentation]
```

but it would be more idiomatic to use filt. Looks like the equivalent would be:

```
pip install flit
flit install --extras testing documentation
```

(untested)

Also it looks like the Github Actions CI builds are passing, but again it would probably be nicer to invoke flit than pip here

- https://github.com/wagtail/wagtail-localize/blob/ebde7a3b56fe46be428b00b74cb08b7be948ed4d/.github/workflows/test.yml#L159
- https://github.com/wagtail/wagtail-localize/blob/ebde7a3b56fe46be428b00b74cb08b7be948ed4d/.circleci/config.yml#L59

and we're going to want to create our cache keys based on hashing `pyproject.toml`, not `setup.py`

- https://github.com/wagtail/wagtail-localize/blob/ebde7a3b56fe46be428b00b74cb08b7be948ed4d/.github/workflows/test.yml#L64
- https://github.com/wagtail/wagtail-localize/blob/ebde7a3b56fe46be428b00b74cb08b7be948ed4d/.github/workflows/test.yml#L135

Can you add those to the list of next things to look at.

As I say, I'll have a more detailed look next week.

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+[tool.flit.module]
+name = "wagtail_localize"
+
+[tool.flit.sdist]
+exclude = [
+    "wagtail_localize/static_src",
+    "wagtail_localize/test",
+    "wagtail_localize/tests",
+    "wagtail_localize/machine_translators/tests",
+    "wagtail_localize/segments/tests",
```
chris48s, line 71: Could we exclude the tests with a glob pattern instead of individually listing each app here?

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "*.js",
+    "*.json",
+    "*.ini",
+    "*.yml",
+    "netlify.toml",
+    "testmanage.py",
+]
+include = [
+    "wagtail_localize/static"
+]
```
chris48s, line 86: This comment is probably more a clarification for @zerolab  when he comes to review it than anything else as it confused me initially:

The reason why we have to explicitly include this here (but we _don't_ have to do this for the templates or .mo files, etc) is because flit respects `.gitignore`, so if we want to include files in the package build that are not checked into source control we have to tell flit to do it.

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "Programming Language :: Python :: 3.10",
+    "Framework :: Django",
+    "Framework :: Django :: 2.2",
+    "Framework :: Django :: 3.2",
+    "Framework :: Django :: 4.0",
+    "Framework :: Wagtail",
+    "Framework :: Wagtail :: 2",
+    "Framework :: Wagtail :: 3"
+]
+dynamic = ["version"]
```
chris48s, line 30: Something to note. Tbh this is probably another note for when @zerolab  comes to review it rather than something actionable right now: Using this configuration, flit gets the version when it builds the package from the project's `__version__` variable. Broadly, this is _a good thing_. In this specific package, our `__init__.py` has a `from django...` import in it: https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/__init__.py which means that in order to read the value of `__version__`, django must be installed in whatever environment we are running `flit build` in. i.e: we can't just run `flit build`. We've got to install at least the `production` dependencies to build the package. That's not necessarily the end of the world but just thought I'd flag it as a possible gotcha.. the existing `setup.py` actually reads it straight from `version.py`, presumably for this reason?

chris48s (commented):
Thanks for your work on this so far Victor. Its looking good. I've spent quite a while looking over this as we are both learning as we go here. One of the things I did as part of testing this was I installed it as a local package inside a wagtail bakery instance as a host app. No comments on that - works great but I wanted to confirm that I've tested that particular use-case.

pyproject.toml
zerolab: that is correct.

chris48s: OK. Do you view the fact we'd have to install django before we can build the package as a showstopper for this approach, something we need to work around, or just something we need to bear in mind?

zerolab: 🤔 I would like things to be as simple as possible.
If there's a good way of working around the `__init__.py` import I'd be very much up for it

zerolab: > For example, does anything bad actually happen if default_app_config = "wagtail_localize.apps.WagtailLocalizeAppConfig" is set on django >= 3.2, or is it just ignored?
you get a bunch of annoying warnings. so not the end of the world if we just have the warnings until we drop support for Django 3.2

> Presumably you'll need to be able to run ./testmanage.py compilemessages anyway as part of the release process
ah, yes! that is true. so we need a virtual env with the dependencies anyway. So maybe update the makefile with the appropriate steps and then it is a matter of documentation.

thaks for looking at this more in-depth. in the end we have the env complexity anyway /me goes and hides in the shadows again

chris48s: Yeah no worries. As I say, these are the sort of things we only learn by using this stuff in the real world. We won't hit this stuff just reading the docs and making a proof-of-concept.

pyproject.toml
chris48s: Turns out the answer to this is "no". Trying to do it with glob patterns throws

```
Config error: Recursive globbing (**) is not supported yet (in exclude pattern 'wagtail_localize/**/tests')
```

Looks like a PR got merged for this about a week ago https://github.com/pypa/flit/pull/550 so will hopefully be in a release soon.

chris48s:
I don't want this to spill past the end of the rock quarter so I've just pushed the commits to address the review comments myself

pyproject.toml
engineervix: Indeed :slightly_smiling_face:, I added a comment on [9a15c53](https://github.com/engineervix/wagtail-localize/commit/9a15c53c0f05109ffbdf17a69433c49f2883ceea) concerning this

setup.py
```diff
@@ -1,73 +0,0 @@
-#!/usr/bin/env python
```
zerolab, line 1: the one minor downside to this is that we are likely to lose the "[used by](https://github.com/wagtail/wagtail-localize/network/dependents?package_id=UGFja2FnZS00NDI2MDYwNzk%3D)" GitHub feature, but we can live with it

zerolab (approved):
Took forever to get to this. This is excellent. Thank you @engineervix and @chris48s 

A few observations from local testing:

**Package time**

`time make package`

setuptools | flit
-----------|-------------
 `0.38s user 0.72s system 78% cpu 1.405 total` | `0.30s user 0.55s system 88% cpu 0.956 total`


**File sizes**

type  | setuptools | flit    | diff
------|------------|---------|-------
egg   | 272174     | 258246  | 5.12% ⬇️
wheel | 518778     | 458771  | 11.56% ⬇️

<details>
<summary>Wheel diff (excluding package metadata files)</summary>

```bash
deleted:    wagtail_localize/machine_translators/tests/__init__.py
deleted:    wagtail_localize/machine_translators/tests/test_dummy_translator.py
deleted:    wagtail_localize/segments/tests/__init__.py
deleted:    wagtail_localize/segments/tests/test_segment_extraction.py
deleted:    wagtail_localize/segments/tests/test_segment_ingestion.py
deleted:    wagtail_localize/segments/tests/test_segmentvalue.py
deleted:    wagtail_localize/test/__init__.py
deleted:    wagtail_localize/test/apps.py
deleted:    wagtail_localize/test/migrations/0001_initial.py
deleted:    wagtail_localize/test/migrations/__init__.py
deleted:    wagtail_localize/test/models.py
deleted:    wagtail_localize/test/settings.py
deleted:    wagtail_localize/test/urls.py
deleted:    wagtail_localize/test/wagtail_hooks.py
deleted:    wagtail_localize/tests/__init__.py
deleted:    wagtail_localize/tests/test_convert_to_alias.py
deleted:    wagtail_localize/tests/test_edit_translation.py
deleted:    wagtail_localize/tests/test_get_translatable_fields.py
deleted:    wagtail_localize/tests/test_strings.py
deleted:    wagtail_localize/tests/test_submit_translations.py
deleted:    wagtail_localize/tests/test_synctree.py
deleted:    wagtail_localize/tests/test_tasks.py
deleted:    wagtail_localize/tests/test_translation_components.py
deleted:    wagtail_localize/tests/test_translation_model.py
deleted:    wagtail_localize/tests/test_translations_report.py
deleted:    wagtail_localize/tests/test_translationsource_model.py
deleted:    wagtail_localize/tests/test_update_translations.py
deleted:    wagtail_localize/tests/utils.py
```

</details>

pyproject.toml
merwok: FYI the PR was merged

zerolab: it was, but hasn't made it into a release yet

PR #601, opened 2022-08-18
Fix TestSubmitPageTranslationWithComponents for Wagtail 4

gasman:
The markup for form fields with errors has changed substantially in Wagtail 4, and now there's no straightforward way to assert "an error exists on field X" using a substring match on the HTML. Instead, check the form object directly.


zerolab (approved):
Nice one, thank you @gasman!

PR #604, opened 2022-08-29
DeepL offers a free API now

ramiboutas:
It would be wise to use the python-deepl library: https://github.com/DeepLcom/deepl-python


zerolab:
Note: alternative to #505

@ramiboutas any chance you could add a small test for this

wagtail_localize/machine_translators/tests/test_deepl_translator.py
```diff
@@ -0,0 +1,16 @@
...
+class TestDeeplTranslator(TestCase):
+    def setUp(self):
+        self.free_auth_key = "asd-23-ssd-243-adsf-dummy-auth-key:fx"
+        self.paid_auth_key = "asd-23-ssd-243-adsf-dummy-auth-key:bla"
+    
+    def test_api_free(self):
+        self.assertEqual(auth_key_is_free_account(self.free_auth_key), True)
+    
+    def test_api_paid(self):
+        self.assertEqual(auth_key_is_free_account(self.paid_auth_key), False)
```
zerolab, line 16: This is a start, but I was thinking checking that `DeepLTranslator.translate` uses the correct URL based on the key, since that is what we want to test first and foremost

ramiboutas: That it is what I also wanted but I dont know how to do that; my apologies. I will learn more about writting tests and come back to the issue again.

zerolab: Look for mocking (e.g. https://bhch.github.io/posts/2017/09/python-testing-how-to-mock-requests-during-tests/)

alternatively, move the logic to return a URL to a method in `DeepLTranslator` (e.g. `get_api_endpoint`), and use that in `requests.post`. Then you can test with `override_settings` to change the key

in your test you could call `DeepLTranslator({}).get_api_endpoint()` and check its value

zerolab:
Tidied up and merged in 605c124fccfa6a30e853e88f8b836a09f0cec71e.

Thanks @ramiboutas

PR #621, opened 2022-09-23
Fix string extraction from nullable `RichTextField`

### LINKED ISSUE #487: Error while translating: TypeError: object of type 'NoneType' has no len()
Hello, 
I've faced an issue when I tried to translate a page, specifically:

```
File "/app/.heroku/venv/lib/python3.6/site-packages/wagtail_localize/segments/extract.py", line 183, in extract_segments
    template, strings = extract_strings(field.value_from_object(instance))
File "/app/.heroku/venv/lib/python3.6/site-packages/wagtail_localize/strings.py", line 297, in extract_strings
    soup = BeautifulSoup(html, "html.parser")
File "/app/.heroku/venv/lib/python3.6/site-packages/bs4/__init__.py", line 310, in __init__
    elif len(markup) <= 256 and (
TypeError: object of type 'NoneType' has no len()`
```

When I edited the page and added a value to some random field, published and tried to translate one again - it worked. 
But the issue is definitely something worth to check.

zerolab:
Fixes #487, and replaces #523

Kudos to @vladox and @Timusan


wagtail_localize/test/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.0.5 on 2022-06-05 11:41
+# Generated by Django 4.1.1 on 2022-09-23 14:13
```
chris48s, line 1: This isn't a problem with this PR (if it was actually causing a problem, a test would have failed), but just while I am reading this line, a quick note.

For django packages, the recommended practice is to always generate your migrations with the lowest version of django your package is compatible with (which in this case would be 3.2). See https://docs.djangoproject.com/en/4.1/topics/migrations/#supporting-multiple-django-versions for explanation.

chris48s (approved):

wagtail_localize/test/migrations/0001_initial.py
chris48s: (this matters even less if the migration is for the test app :smile: )

zerolab: it doesn't matter as I've made the necessary modifications for the initial migration manually

vladox:
@zerolab  great job thxs!

PR #622, opened 2022-09-27
Add method to delete translation data on translatable model deletion

### LINKED ISSUE #402: Deleting translated pages/snippets/etc does not delete their data

zerolab:
Changes the default behaviour so that it removes related localize data on translatable model deletion, rather than marking the `Translation` as disabled.

Adds a setting (`WAGTAILLOCALIZE_DISABLE_ON_DELETE` to allow for the old behaviour should it be needed.

Fixes #402

To-Do:

- [x] Add tests
- [x] Update documentation

### DIFF
--- a/wagtail_localize/models.py
+++ b/wagtail_localize/models.py
@@ -1038,8 +1038,8 @@ class Translation(models.Model):
 
     An instance of this model is created whenever an object is submitted for translation into a new language.
 
-    They can be disabled at any time, and are disabled automatically if either the source or destination object is
-    deleted.
+    They can be disabled at any time, and are deleted or disabled automatically if either the source or
+    destination object is deleted.
 
     If the translation of a page is disabled, the page editor of the translation would return to the normal Wagtail
     editor.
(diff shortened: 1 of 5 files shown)


wagtail_localize/models.py
```diff
@@ -2191,14 +2191,59 @@ def disable_translation_on_delete(instance, **kwargs):
...
+    StringTranslation.objects.filter(
+        context__object_id=translation_key, locale=locale_id
+    ).delete()
+    SegmentOverride.objects.filter(
+        context__object__translation_key=translation_key, locale=locale_id
+    ).delete()
+    Translation.objects.filter(source__object_id=translation_key).filter(
+        # Remove the translations where this object was the source
+        Q(source__locale_id=instance.locale_id)
+        # Removed the translations where this object was the destination
```
tbrlpld, line 2212: Just a small typo.

```suggestion
        # Remove the translations where this object was the destination
```

wagtail_localize/models.py
```diff
@@ -2191,14 +2191,59 @@ def disable_translation_on_delete(instance, **kwargs):
...
+        | Q(target_locale_id=instance.locale_id)
```
tbrlpld, line 2213: Can you add some context why we are using `instand.locale_id` instead of `locale_id` here?

wagtail_localize/tests/test_translation_model.py
```diff
@@ -705,9 +709,11 @@ def test_uuid_snippet_save_target(self):
         self.assertEqual(translated_snippet.field.charfield, "Some Test")
 
 
+@override_settings(WAGTAILLOCALIZE_DISABLE_ON_DELETE=True)
 class TestDeleteSourceDisablesTranslation(TestCase):
-    def setUp(self):
-        self.fr_locale = Locale.objects.create(language_code="fr")
+    @classmethod
+    def setUpTestData(cls):
+        cls.fr_locale = Locale.objects.create(language_code="fr")
```
tbrlpld, line 716: Kudos for test speed 😄

wagtail_localize/models.py
zerolab: good spot. did a refactor and missed those

zerolab:
@kaedroho will merge this tomorrow mid-morning, unless I hear otherwise

Issue #615, opened 2022-09-19
Migration fails upgrading to 1.3.0-alpha


iscilyas:
I've been waiting for a wagtail 4.0-compatible wagtail-localize, so I guess I'm one of the earliest adopters...

After upgrading django/wagtail/wagtail-localize (from 1.2.1) I'm seeing the following error when running `makemigrations`:

```
$ ./manage.py makemigrations
Traceback (most recent call last):
  File "/var/www/html/imrozrum.k12.tr/wagtail/./manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 440, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/core/management/base.py", line 402, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/core/management/base.py", line 96, in wrapped
    res = handle_func(*args, **kwargs)
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/core/management/commands/makemigrations.py", line 141, in handle
    loader.check_consistent_history(connection)
  File "/var/www/html/imrozrum.k12.tr/wagtail/venv/lib/python3.10/site-packages/django/db/migrations/loader.py", line 327, in check_consistent_history
    raise InconsistentMigrationHistory(
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration wagtail_localize.0001_initial is applied before its dependency wagtailcore.0076_modellogentry_revision on database 'default'.
```
Anyone else seeing this?

zerolab:
Thanks for the report @iscilyas, this is an interesting issue.
[...]

PR #630, opened 2022-10-07
Fix Revision model change in Wagtail 4.0

### LINKED ISSUE #629: Migration fail installing wagtail-localize fresh on a wagtail 4.1 project
I was trying to add 1.3a4 to a project that I had already upgraded to wagtail 4, and end up with 
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration wagtailcore.0059_apply_collection_ordering is applied before its dependency wagtail_localize.0001_initial

If I understand this correctly the run_before line in the 0001 migration that was added in #616 will cause a failure for everyone trying to install wagtail-localize after they have already run migrate 0059 (I am at 0076 at the moment)

zerolab:
this should cover:
- adding localize to a Wagtail 4 project
- upgrading a project using localize to Wagtail 4

Fixes #629


zerolab:
@ArnarTumi any chance you could test this on your projects `pip install git+https://github.com/wagtail/wagtail-localize.git@fix/pagerevision-revision-switcharoo-dance`. Note that installing it this way won't have the compiled assets so the edit translation view will error. Should not matter as you're testing adding localize to an existing project. If you have any on pre 4.0, would be nice to try it on them too

cc @iscilyas in the context of #615 too

zerolab:
@gasman will merge this as is tomorrow mid-morning, unless I hear otherwise

gasman (approved):
LGTM! This looks like a very neat way of dealing with the unpredictability of the model name. I look forward to stealing it for wagtail-review :-)

zerolab:
Thank you! it took quite some trial and error. Lots of TILs in the process. I am both in awe and somewhat mystified by the power of Django migrations

PR #685, opened 2023-03-30
Use the url field for the project homepage in PyPI

thibaudcolas:
Spotted while doing data analysis of Wagtail packages. [`url`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#url) is specifically meant for the project "Homepage", so is more appropriate than adding a `project_urls` set to "Home".

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "flit_core.buildapi"
 [project]
 name = "wagtail-localize"
 authors = [{name = "Karl Hobley", email = "karl@torchbox.com"}]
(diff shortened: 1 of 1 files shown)


zerolab (approved):

zerolab:
@thibaudcolas I discovered the hard way, that this is not the way when using pyproject.toml

see https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#urls

thibaudcolas:
🤦 sorry! I wouldn’t even have considered that the possible metadata was different between different build tools

PR #697, opened 2023-05-23
WIP: Upgrade for Wagtail 5.0

nickmoreton:
**This work updates the package testing and upgrade considerations for working with Wagtail v5.0**

Also:
- drops support for Django 4.0 (https://docs.wagtail.org/en/stable/releases/5.0.html#dropped-support-for-django-4-0)
- drops support for python 3.7 (end of life is June 2023)

Notable changes from the upgrade considerations:
- Longrunning button class (https://docs.wagtail.org/en/stable/releases/5.0.html#progress-button-button-longrunning-now-relies-on-data-attributes)
- Deprecated icons (https://docs.wagtail.org/en/stable/releases/5.0.html#deprecated-icon-font)


wagtail_localize/modeladmin/templates/wagtail_localize/modeladmin/translatable_edit.html
```diff
@@ -7,7 +7,7 @@
 
 {% block form_actions %}
   <div class="dropdown dropup dropdown-button match-width">
-    <button type="submit" class="button action-save button-longrunning" data-clicked-text="{% trans 'Saving…' %}">
+    <button type="submit" class="button action-save button-longrunning" data-controller="w-progress" data-action="w-progress#activate" data-w-progress-active-value="{% trans 'Saving…' %}">
```
zerolab, line 10: have you tested this in Wagtail 4.1? the `data-controller` change only happens in Wagtail 5.0

wagtail_localize/static_src/editor/components/TranslationEditor/footer.tsx
```diff
@@ -103,7 +103,7 @@ const EditorFooter: FunctionComponent<EditorProps> = ({
     if (perms.canUnpublish) {
         actions.push(
             <a className="button action-secondary" href={links.unpublishUrl}>
-                <Icon name="download-alt" />
+                <Icon name="download" />
```
zerolab, line 106: Doesn't work with Wagtail < 5.0. Ditto for all icon changes.
The solution would be to add the old icons and register them for Wagtail < 5.0

wagtail_localize/static_src/editor/components/TranslationEditor/footer.tsx
```diff
@@ -124,7 +124,9 @@ const EditorFooter: FunctionComponent<EditorProps> = ({
                     name="action"
                     value="publish"
                     className="button button-longrunning "
-                    data-clicked-text={gettext('Publishing...')}
```
zerolab, line 127: needs to stay for Wagtail 4.1 compat

zerolab (changes requested):
Thank you for making a start, @nickmoreton
There's more work needed. It is worth testing such upgrades with the new Wagtail version, but they always must be tested with the supported versions (usually LTS, and whatever version is before the one you're doing the upgrade).

Unlike sites, packages have to take in consideration multiple Wagtail versions

nickmoreton:
> There's more work needed. It is worth testing such upgrades with the new Wagtail version, but they always must be tested with the supported versions (usually LTS, and whatever version is before the one you're doing the upgrade).

Thanks @zerolab How did I not think of that, I've done enough of these 🥲 I'll look again.

zerolab:
Continued in #699

wagtail_localize/static_src/editor/components/TranslationEditor/footer.tsx
zerolab: looks like I was hasty. 4.0 added all the correctly named icons, so this was good

PR #711, opened 2023-09-06
Async race condition fix - Moved publishing to transaction post-commit

Abdul-Dridi:
Our wagtail setup has us processing wagtail signals asynchronously (in celery tasks), we've noticed that when publishing localised content, the tasks asynchronous page database query oftentimes retrieves the page state before the translations have been synced.

From our testing, this occurs when a page db query is made before the translation transaction has been committed.

This pr pushes the page revision publish to be executed after the transaction has been successfully [committed](https://docs.djangoproject.com/en/4.2/topics/db/transactions/#django.db.transaction.on_commit).


wagtail_localize/tests/test_edit_translation.py
```diff
@@ -89,82 +92,84 @@
...
-                test_slugfield="a-slug-field",
-                test_urlfield="https://www.example.com",
-                test_richtextfield=RICH_TEXT_DATA,
-                test_streamfield=StreamValue(
-                    TestPage.test_streamfield.field.stream_block,
-                    STREAM_DATA,
-                    is_lazy=True,
-                ),
-                test_snippet=self.snippet,
+        with patch.object(transaction, "on_commit", side_effect=lambda func: func()):
```
zerolab, line 95: I really don't think these setup tasks need to be wrapped like this.
we are only interested on the create/update translation tests to issue a on_commit

Abdul-Dridi: Yeah that's a good point, moved the patch to an overriden setup method in the relevant test class. (I'm wrapping the setup task because the Translation.save_target method used executes my changed code)

Unless I'm missing something, the only way to bring the patch into the test_edit_page_translation method itself would be to split up the setup method and call the relevant part of it in the actual test

zerolab:
Thank you for this @Abdul-Dridi, tidied up, brought up to date with `main` and merged in #720

PR #716, opened 2023-09-23
Locale filter selector based on installed languages

jhonatan-lopes:
## The problem

On the "Translations" report, the current way to filter by locales is to type the language code in the field exactly as it was defined by the developers. This causes some issues.

First, we could have some editors/admins that might just want a quick reference into translations and might not experts in the field. Hence, it is not fair to expect them to know the language codes for all installed languages. For instance, although ubiquitous, one might not know that "de" is the language code for the German language.

The second problem is more subtle and more difficult to debug when it happens. If a project defines their `WAGTAIL_CONTENT_LANGUAGES` as

```python
LANGUAGE_CODE = "en"
WAGTAIL_CONTENT_LANGUAGES = LANGUAGES = (
    ("en", gettext_lazy("English")),
    ("de", gettext_lazy("German")),
    ("es", gettext_lazy("Spanish")),
    ("fr", gettext_lazy("French")),
    ("fy-NL", gettext_lazy("Frisian")),
    ("nl", gettext_lazy("Dutch")),
    ("pl", gettext_lazy("Polish")),
    ("pt-BR", gettext_lazy("Portuguese (Brazil)")),
    ("sw", gettext_lazy("Swahili")),
)
```
for example, the field as it current stands will only recognise "pt-BR" as the "Portuguese (Brazil)" language, but not "pt-br".

From the [ITEF wiki](https://en.wikipedia.org/wiki/IETF_language_tag):

> Subtags are not [case-sensitive](https://en.wikipedia.org/wiki/Case-sensitive), but the specification recommends using the same case as in the Language Subtag Registry, where region subtags are [UPPERCASE](https://en.wikipedia.org/wiki/UPPERCASE), script subtags are [Title Case](https://en.wikipedia.org/wiki/Title_case), and all other subtags are [lowercase](https://en.wikipedia.org/wiki/Lowercase). This capitalization follows the recommendations of the underlying ISO standards.

So that the "correct-er" form would be "pt-BR", although if subtags are supposed to not be case-senstive then both cases must filter properly (which is not the currently behaviour).

## The solution

This PR leverages the language codes defined in `WAGTAIL_CONTENT_LANGUAGES` to provide chooser fields for the locales instead of relying on having to type the proper language code. This abstracts the language code information from editors/admins with the bonus of displaying language names localized to their Wagtail admin choice.

An "All" option is defined to display translations in all locales.

#### Screenshots:

Using the provided test app with:

```python
LANGUAGES = WAGTAIL_CONTENT_LANGUAGES = [
    ("en", "English"),
    ("fr", "French"),
    ("fr-CA", "French (Canada)"),
    ("es", "Spanish"),
]
```

image

image

image

### DIFF
--- a/wagtail_localize/tests/test_translations_report.py
+++ b/wagtail_localize/tests/test_translations_report.py
@@ -135,6 +135,17 @@ class TestTranslationReport(TestCase, WagtailTestUtils):
         self.assertNotIn(self.blog_index_translation, response.context["object_list"])
         self.assertNotIn(self.blog_post_translation, response.context["object_list"])
 
+    def test_filter_by_source_locale_all_options(self):
+        response = self.client.get(
+            reverse("wagtail_localize:translations_report") + "?source_locale=all"
+        )
+
+        self.assertIn(self.snippet_translation, response.context["object_list"])
+        self.assertIn(self.homepage_translation, response.context["object_list"])
+        self.assertIn(self.de_homepage_translation, response.context["object_list"])
+        self.assertIn(self.blog_index_translation, response.context["object_list"])
+        self.assertIn(self.blog_post_translation, response.context["object_list"])
+
     def test_filter_by_target_locale(self):
         response = self.client.get(
             reverse("wagtail_localize:translations_report") + "?target_locale=de"
@@ -145,3 +156,14 @@ class TestTranslationReport(TestCase, WagtailTestUtils):
         self.assertIn(self.de_homepage_translation, response.context["object_list"])
         self.assertNotIn(self.blog_index_translation, response.context["object_list"])
         self.assertNotIn(self.blog_post_translation, response.context["object_list"])
+
+    def test_filter_by_target_locale_all_options(self):
+        response = self.client.get(
+            reverse("wagtail_localize:translations_report") + "?target_locale=all"
+        )
+
+        self.assertIn(self.snippet_translation, response.context["object_list"])
+        self.assertIn(self.homepage_translation, response.context["object_list"])
+        self.assertIn(self.de_homepage_translation, response.context["object_list"])
+        self.assertIn(self.blog_index_translation, response.context["object_list"])
+        self.assertIn(self.blog_post_translation, response.context["object_list"])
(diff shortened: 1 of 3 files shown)


wagtail_localize/views/report.py
```diff
@@ -111,8 +120,19 @@ class TranslationsReportFilterSet(WagtailFilterSet):
         field_name="source__specific_content_type", label=gettext_lazy("Content type")
     )
     source_title = SourceTitleFilter(label=gettext_lazy("Source title"))
-    source_locale = LocaleFilter(field_name="source__locale")
-    target_locale = LocaleFilter()
+    source_locale = LocaleFilter(
+        field_name="source__locale",
+        choices=_get_locale_choices,
+        empty_label=None,
+        null_label="All",
```
zerolab, line 127: needs to be translated

wagtail_localize/views/report.py
```diff
@@ -98,9 +99,17 @@ def filter(self, qs, value):
         )
 
 
-class LocaleFilter(django_filters.CharFilter):
+def _get_locale_choices():
+    choices = [
+        (language_code, display_name)
+        for language_code, display_name in get_content_languages().items()
+    ]
+    return choices
```
zerolab, line 107:
```suggestion
    return list(get_content_languages().items())
```

wagtail_localize/tests/test_translations_report.py
```diff
@@ -135,6 +135,17 @@ def test_filter_by_source_locale(self):
         self.assertNotIn(self.blog_index_translation, response.context["object_list"])
         self.assertNotIn(self.blog_post_translation, response.context["object_list"])
 
+    def test_filter_by_source_locale_all_options(self):
```
zerolab, line 138: would be good to test that the locale filter gives you the `get_content_languages` choices

zerolab (changes requested):
This is a welcome improvement @jhonatan-lopes

Left a few notes

wagtail_localize/tests/test_translations_report.py
jhonatan-lopes: Done

wagtail_localize/views/report.py
jhonatan-lopes: Nice :+1:

wagtail_localize/views/report.py
jhonatan-lopes: My bad, I forgot that :facepalm: 

Do I need to translate anything from `get_content_language` as well? My assumption is that this must be handled from Wagtail/settings side

zerolab: considering get_content_language pull stuff from settings, that is up to the implementer, not us

zerolab (approved):

PR #739, opened 2023-11-07
Fix splitting of en locale codes according to DeepL requierements

### LINKED ISSUE #738: DeepL target_lang for EN-GB and EN-US
The [DeepL API documentation](https://www.deepl.com/de/docs-api/translate-text/translate-text) specifies to use EN-GB or EN-US to archieve the desired translation, however following code in the deepl.py will chop off the -GB or -US and the deprecated unspecified route will be used.

```python
def language_code(code, is_target=False):
    # DeepL supports targeting Brazillian Portuguese but doesn't have this for other languages
    if is_target and code in ["pt-pt", "pt-br"]:
        return code
    return code.split("-")[0].upper()
```

I suggest to add en-gb and en-us to the array inside the if statement above to respect users desired locale setting.
Please tell me if this is shortsightet or more complicated then this 😅

unreadableusername:
Added American and British English locales to translation target locale array. 
fixes #738

### DIFF
--- a/wagtail_localize/machine_translators/deepl.py
+++ b/wagtail_localize/machine_translators/deepl.py
@@ -6,8 +6,8 @@ from .base import BaseMachineTranslator
 
 
 def language_code(code, is_target=False):
-    # DeepL supports targeting Brazillian Portuguese but doesn't have this for other languages
-    if is_target and code in ["pt-pt", "pt-br"]:
+    # DeepL supports targeting Brazillian Portuguese and requieres to specifically request American or British English.
+    if is_target and code in ["pt-pt", "pt-br", "en-us", "en-gb"]:
         return code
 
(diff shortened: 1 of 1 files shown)


unreadableusername:
I see the cheks fail, but judging from the output this has nothing to do with my small change, but rather with some dev dependencies that might need a bump right? I am just anxious because this is my first contribution 🥳

zerolab:
We have some Netlify jobs that fail and are on the list to be replaced. Triggered the actual CI tasks (they require approval to be run on PRs from first time contributors)

zerolab:
Added tests and merged in https://github.com/wagtail/wagtail-localize/commit/a3efa8f65de3445c75ea271b1b0ee513a445deaa
Thank you

PR #751, opened 2023-12-16
Add support for saving revisions of snippets

### LINKED ISSUE #666: Support for translating snippets in draft mode
Since [Wagtail 4.0 snippets can be saved as draft](https://docs.wagtail.org/en/stable/topics/snippets.html#saving-draft-changes-of-snippets) by using DraftStateMixin.
When translating a snippet which is still in draft mode the following exception is raised:

```
  File "/Users/herbert/Library/Caches/pypoetry/virtualenvs/sportpark-backend-ztpaQDXI-py3.10/lib/python3.10/site-packages/wagtail_localize/operations.py", line 90, in translate_object
    translator.create_translations(instance)
  File "/Users/herbert/Library/Caches/pypoetry/virtualenvs/sportpark-backend-ztpaQDXI-py3.10/lib/python3.10/site-packages/wagtail_localize/operations.py", line 79, in create_translations
    translation.save_target(user=self.user, publish=publish)
  File "/Users/herbert/Library/Caches/pypoetry/virtualenvs/sportpark-backend-ztpaQDXI-py3.10/lib/python3.10/site-packages/wagtail_localize/models.py", line 1334, in save_target
    self.source.create_or_update_translation(
  File "/Users/herbert/Library/Caches/pypoetry/virtualenvs/sportpark-backend-ztpaQDXI-py3.10/lib/python3.10/site-packages/wagtail_localize/models.py", line 742, in create_or_update_translation
    raise CannotSaveDraftError
wagtail_localize.models.CannotSaveDraftError
```

I guess this check was implemented before snippets could be saved as draft, and can be simply removed? 🤔 or using `isinstance(original, DraftStateMixin)` instead of `isinstance(original, Page)`?

https://github.com/wagtail/wagtail-localize/blame/abf2853ad7660dde7b9010de36187e618d4d9184/wagtail_localize/models.py#L740-L742

mcmeeking:
I'm working on a project which has a need for setting new snippet translations as drafts by default, so I wanted to close #666.

Changes:

- `page_revision` field in `TranslationLog` model is now simply `revision`.
- Default action when `WAGTAILLOCALIZE_SYNC_LIVE_STATUS_ON_TRANSLATE = False` when creating/updating translations for a `DraftStateMixin` model is to save as a draft.
- Publishing translated `DraftStateMixin` models publishes the revision in the same way as publishing a translated `Page` model (using `save_target()` rather than just copying the segments).


wagtail_localize/models.py
```diff
@@ -804,8 +804,19 @@ def create_or_update_translation(
                     if publish:
                         transaction.on_commit(page_revision.publish)
 
+                elif isinstance(translation, DraftStateMixin):
+                    # We copied another instance which may be live, so we make sure this one matches the desired state
+                    translation.live = publish
+                    translation.save()
+
+                    # Create a new revision of the Snippet
+                    page_revision = translation.save_revision(user=user)
```
zerolab, line 813: nitpick: this is not a page revision anymore

wagtail_localize/test/models.py
```diff
@@ -45,6 +51,23 @@ class TestSnippet(TranslatableMixin, ClusterableModel):
     ]
 
 
+@register_snippet
+class TestSnippetNoDraft(TranslatableMixin, ClusterableModel):
+    field = models.TextField(gettext_lazy("field"))
+    # To test field level validation of snippets
+    small_charfield = models.CharField(max_length=10, blank=True)
```
zerolab, line 58: I don't think we need this as the whole purpose for it is to test that we cannot save drafts.

I would go as far as making it a regular model (i.e. no `@register_snippet`). Snippets are supercharged Django models with an aim to replicate modeladmin in the Wagtail admin. 

Something like:

```python
class TestNoDraftModel(TranslatableMixin):
    field = models.CharField(max_length=10, blank=True)

    translatable_fields = [
        TranslatableField("field"),
    ]

    panels = [
        FieldPanel("field"),
    ]
```

should be plenty

wagtail_localize/tests/test_edit_translation.py
```diff
@@ -2128,7 +2128,7 @@ def test_publish_snippet_translation(self):
         log = TranslationLog.objects.get()
         self.assertEqual(log.source, self.snippet_source)
         self.assertEqual(log.locale, self.fr_locale)
-        self.assertIsNone(log.page_revision)
+        self.assertIsInstance(log.page_revision, Revision)
```
zerolab, line 2131: note to self: this is correct (in that TranslationLog does have a `page_revision` FK - https://github.com/wagtail/wagtail-localize/blob/a3efa8f65de3445c75ea271b1b0ee513a445deaa/wagtail_localize/models.py#L1368 - this used to point to PageRevision, but that changed. One for me to address

zerolab: Having said that, we do need a test for `TestSnippet` and drafts / publishing

zerolab (changes requested):
Hey @mcmeeking,

This is a great start! I think making `TestSnippet` inherit from `DraftStateMixin` was a good idea. Made a suggestion to improve the "no draft" model, though.

We need a few tests with drafts / revisions for the test snippet, so as to cover the new logic paths. 

And finally, https://www.wagtail-localize.org/how-to/installation/#disabling-default-publication-of-translated-pages could use an update as that is no longer entirely true

wagtail_localize/models.py
mcmeeking: I've updated this (and the `TranslationLog` model field) to simply "revision", since it's just an FK to Revision rather than PageRevision.

wagtail_localize/tests/test_edit_translation.py
mcmeeking: I've added a test to ensure draft state Snippets can be published in the same way as Pages (by `Translation.save_target()`)

wagtail_localize/tests/test_translation_model.py
```diff
@@ -699,6 +700,71 @@ def test_save_target_cant_save_snippet_as_draft(self):
...
+
+        french_snippet = snippet.get_translation(self.fr_locale)
+
+        self.assertTrue(french_snippet.live)
+        self.assertEqual(french_snippet.field, "Contenu de test")
+
+        log = TranslationLog.objects.get()
+        self.assertEqual(log.source, source)
+        self.assertEqual(log.locale, self.fr_locale)
+        self.assertIsInstance(log.revision, Revision)
```
mcmeeking, line 766: @zerolab, just a note on this:

This checks that the logged revision is a Revision instance, but does not check that the log.revision matches `self.fr_snippet.live_revision` which is probably a better test. The reasoning for this is that the `self.fr_snippet.live_revision` is `None` when called here, and my intuition is that this is because the `publish()` method is only called `on_commit` in the `save_target` method.

If this sounds correct to you as well, then the current assertions should be fine, but let me know if you think this is incorrect as it may mean that the publish actions are not successfully running.

zerolab: @mcmeeking thank you for adding tests. Will aim to get it reviewed in the next few days (and give it the proper attention it deserves)

docs/how-to/installation.md
```diff
@@ -54,10 +54,13 @@ class MyPage(Page):
     # ...
 ```
 
-## Disabling default publication of translated pages
+## Saving new translations as drafts
 
-Live pages that are submitted for translation are made live immediately. If you wish live pages submitted for
-translation to remain as drafts, set `WAGTAILLOCALIZE_SYNC_LIVE_STATUS_ON_TRANSLATE = False` in your settings file.
+Live draft-able models (i.e. subclasses of `Page`, and models which inherit `DraftStateMixin` and `RevisionMixin`)
```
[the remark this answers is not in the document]
mcmeeking: Added in https://github.com/wagtail/wagtail-localize/pull/751/commits/8c2253f60936b1234ddbe93620ac7f7683e03d95

wagtail_localize/tests/test_translation_model.py
mcmeeking: Thanks for this, I checked how the page was being tested for this and noticed the `_mock_on_commit` being passed by the patch decorator so I've updated the test to check `live_revision` now.

wagtail_localize/operations.py
```diff
@@ -33,7 +33,7 @@ def create_translations(self, instance, include_related_objects=True):
             return
         self.seen_objects.add(instance.translation_key)
 
-        source, created = TranslationSource.get_or_create_from_instance(instance)
+        source, _ = TranslationSource.get_or_create_from_instance(instance)
```
zerolab, line 36: I'd suggest using `_created` to avoid a scenario where we import gettext/gettext_lazy as _

mcmeeking: Good point, reverted in the latest commit 👍

mcmeeking:
@zerolab - I've just taken this for a test drive today and it's not right yet, the snippet edit view isn't rendering correctly - I tried converting this to a draft but was not able to. This is just an FYI not to merge this until further changes have been pushed.

mcmeeking:
> @zerolab - I've just taken this for a test drive today and it's not right yet, the snippet edit view isn't rendering correctly. This is just an FYI not to merge this until further changes have been pushed.

I'm pretty sure this is actually fine, I'm just not sure how to build the JS components when including my fork as a requirements.txt target, so the React components aren't loading on the edit page because `/wagtail_localize/js/wagtail-localize.js` is returning a 404 for me (since it's not being built).

zerolab:
you can run `nvm use`, `npm ci`, `npm run build`, then `pip install -e path/to/local/wagtail-localize` locally and you should get all the assets

zerolab:
@mcmeeking the changes so far look quite good. However the translation editing interface doesn't pick up the drafts.

You can test this locally by:

1. in your repository root run ([`fnm use`](https://github.com/Schniz/fnm) or `nvm use`)
2. then `npm ci` and `npm run build`
3. edit `tox.ini` so the `[testenv:interactive]` section looks like:

```
[testenv:interactive]
basepython = python3.11

deps =
    Wagtail < 6.0

comands_pre =
# the rest of the current tox.ini below
```

4. (`pip install tox` if you haven't already) and `tox -e interactive`
5. go to `http://localhost:8020/admin` (admin/changeme)
6. then Snippets > Test snippets and save a draft and play around

mcmeeking:
Thanks for this, I'll try and take a look and sort this week 👍

zerolab:
I will do some more tests tomorrow (and compare with the page functionality)

zerolab:
Having compared to what happens with pages, I think this does what it says on the tin.
Will re-test with this lens tonight, and likely open a follow up issue to at least document what it means when you have a published page/draftable model with a translation, then you save a draft and sync translation

mcmeeking:
> @mcmeeking the changes so far look quite good. However the translation editing interface doesn't pick up the drafts.

Sorry, took a little longer to get around to this than expected - I've taken it for another test drive yesterday following the steps suggested and it appears to be saving draft and allowing publish fine - but I think I see what you're referring to about the UI in that the status in the side panel still states "Live" even when there are new drafts.

From my checks, this seems to be consistent with the Page translation edit page too.

1. Is that what you were referring to?
2. Shall I try and fix that in this PR as well?

zerolab:
Merged in 2773147 (with the removal of the 4.2 conditional

Thank you

PR #747, opened 2023-12-07
Change dependencies to allow Django 5.0.

ACK1D:
Add tox envs for combos of Python with Django 5.0 and Wagtail 5.2.2.

PR #754, opened 2023-12-20
updating tests as .assertFormError() is removed in Django 5

softquantum:


softquantum:
I tested with tox locally and all tests passed, why is the test deployment failing ?

zerolab:
Those are Netlify errors (on the list to move away from hosting docs there)
Just triggered the proper CI jobs.

Also, any chance you could fold in the changes from #747? So as to actually test with Django 5.0

softquantum:
> Those are Netlify errors (on the list to move away from hosting docs there) Just triggered the proper CI jobs.
> 
> Also, any chance you could fold in the changes from #747? So as to actually test with Django 5.0

Will check

zerolab:
I am happy to merge this as is, and we can focus on the rest of the errors.
Not sure what's happening there without looking into it, to be honest.

Will try to have a bit of a dig tomorrow morning, unless you beat me to it

softquantum:
Will try to do it before ;)

PR #756, opened 2023-12-22
Prevent Translation Object Duplicates

### LINKED ISSUE #749: Prevent `TranslationCreator` create `Translation` objects with the same locale as the source locale`
From an internal report where a site somehow ended up with multiple `Translation` objects where `target_locale` was the same as `soruce.locale` which the resulted in some odd behaviours.

Update https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/operations.py#L63-L71 to skip any processing when `target_locale == source.locale`

ACK1D:
Prevent the creation of duplicate Translation objects with the same source and target locale. Replaced the explicit check for existing translations with the use of the `get_or_create` method, ensuring atomicity and avoiding potential duplicates. Fix #749

### DIFF
--- a/wagtail_localize/tests/test_operations.py
+++ b/wagtail_localize/tests/test_operations.py
@@ -53,17 +53,6 @@ class TranslationOperationsTest(TestCase):
             user=self.user, target_locales=self.target_locales
         )
 
-        # Check if TranslationSource already exists
-        source, created = TranslationSource.objects.get_or_create(
-            object_id=self.page.translation_key,
-            specific_content_type=ContentType.objects.get_for_model(TestPage),
-            locale=self.page.locale,
-        )
-
-        # If not created, prepare the source
-        if created:
-            prepare_source(source)
-
     def test_create_translations_skips_duplicate(self):
         # Call create_translations() to check that only one translation has been created
         self.translation_creator.create_translations(self.page)
--- a/wagtail_localize/tests/test_operations.py
+++ b/wagtail_localize/tests/test_operations.py
@@ -1,5 +1,4 @@
 from django.contrib.auth import get_user_model
-from django.contrib.contenttypes.models import ContentType
 from django.test import TestCase
 from wagtail.models import Locale, Page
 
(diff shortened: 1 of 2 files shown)


zerolab:
This is a good start @ACK1D. Thank you for correcting the typos too.

Could you please add some tests? I'd add a new `test_operations.py` test file under https://github.com/wagtail/wagtail-localize/tree/main/wagtail_localize/tests, add a secondary locale ([example](https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/tests/test_edit_translation.py#L137)) and a page, then add a test that instantiates 
`TranslationCreator` with the default and new locales for `target_locales` and calls `create_translations()` to check that only one translation has been created

ACK1D:
@zerolab thanks a bunch for the detailed info. Can you check this out?

wagtail_localize/tests/test_operations.py
```diff
@@ -0,0 +1,96 @@
...
+        # Check if TranslationSource already exists
+        source, created = TranslationSource.objects.get_or_create(
+            object_id=self.page.translation_key,
+            specific_content_type=ContentType.objects.get_for_model(TestPage),
+            locale=self.page.locale,
+        )
+
+        # If not created, prepare the source
+        if created:
+            prepare_source(source)
```
zerolab, line 65: I am pretty sure `create_test_page()` will sort this out, so it should not be needed.

Other than that, `create_test_page` and `prepare_source` is a repeat from `/test_translation_model.py` and `test_translationsource_model.py` and could be extracted to their own helpers/utils, but I will sort that out

zerolab (approved):
One non-blocking note, but this is great work, thank you!

wagtail_localize/tests/test_operations.py
ACK1D: Thank you. Fixed.

zerolab (approved):

wagtail_localize/tests/test_operations.py
```diff
@@ -0,0 +1,85 @@
+from django.contrib.auth import get_user_model
+from django.contrib.contenttypes.models import ContentType
```
zerolab, line 2:
```suggestion
```

zerolab:
Merged in 8b32d63. Thanks @ACK1D

PR #762, opened 2024-01-06
Fix the Locale create/edit/delete success messages

ACK1D:
This pull request addresses an issue related to success messages in the CreateView, EditView, and DeleteView for Locales. The previous implementation of success messages was not handling formatting properly, and this PR introduces improvements. Fix #761

PR #763, opened 2024-01-07
Add tests with success messages for locale creation, editing, and deletion

ACK1D:
Added tests for #762 (success messages for create, edit, and delete locale). Some logic has been refactored into BaseLocaleTestCase, and correcting some tests for use it. Please review code and let me know if need some work.

### DIFF
--- a/wagtail_localize/locales/tests.py
+++ b/wagtail_localize/locales/tests.py
@@ -18,67 +18,74 @@ class BaseLocaleTestCase(TestCase, WagtailTestUtils):
         self.login()
         self.english = Locale.objects.get()
 
-    def execute_request(self, method, view_name, *args, **kwargs):
+    def execute_request(
+        self, method, view_name, *args, params=None, post_data=None, **kwargs
+    ):
         # Helper method to execute HTTP requests
         url = reverse(view_name, args=args)
-        params = kwargs.get("params", {})
-        post_data = kwargs.get("post_data", {})
+        params = params or {}
+        post_data = post_data or {}
 
         if method == "GET":
-            return self.client.get(url, params)
+            return self.client.get(url, params, **kwargs)
         elif method == "POST":
-            return self.client.post(url, post_data)
+            return self.client.post(url, post_data, **kwargs)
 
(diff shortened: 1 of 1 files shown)


ACK1D:
Hi @zerolab
I've made some changes to code, and it would be helpful to update the Codecov report. Could you please trigger a new run? Thank you!

zerolab:
@ACK1D thank you very much for this. I am not sure how to make Codecov update its comment, but https://app.codecov.io/gh/wagtail/wagtail-localize/pull/762 shows a +0.11% increase

wagtail_localize/locales/tests.py
```diff
@@ -1,61 +1,124 @@
...
+        response = self.client.get(reverse("wagtaillocales:add"))
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, "wagtaillocales/create.html")
-
         self.assertEqual(
             response.context["form"].fields["language_code"].choices, [("fr", "French")]
         )
 
     def test_create(self):
-        response = self.post(
```
zerolab, line 52: `TestLocaleCreateView` still has a `post` method that posts to `reverse("wagtaillocales:add")`. I'd say keep using it, especially that we're using that in at least one other place

wagtail_localize/locales/tests.py
```diff
@@ -190,10 +277,14 @@ def post(self, post_data=None, locale=None):
         )
 
     def test_simple(self):
-        response = self.get()
+        # Test rendering the edit view with simple data
+        response = self.client.get(
```
zerolab, line 281: we now seem to have a mix of `self.client.get` like here, and `self.get` below (like in `test_invalid_language`). It would be nice to be consistent

zerolab (changes requested):
@ACK1D kudos for expanding the test locales test coverage.
I left a few notes around the consistency of self.client.get/post vs self.get/post which would be nice to tidy up and be consistent about.

zerolab (approved):
Done is better than perfect. Thanks @ACK1D for the tweaks

PR #779, opened 2024-02-19
Add formality support to DeepL translator

Abdul-Dridi:
This PR adds formality support to the existing DeepL translator (https://www.deepl.com/docs-api/translate-text). This is done by specifying an optional option.

### DIFF
--- a/docs/how-to/integrations/machine-translation.md
+++ b/docs/how-to/integrations/machine-translation.md
@@ -124,11 +124,13 @@ WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
     "CLASS": "wagtail_localize.machine_translators.deepl.DeepLTranslator",
     "OPTIONS": {
         "AUTH_KEY": "<Your DeepL key here>",
-        # Optional DeepL API option
+        # Optional DeepL API option. Accepts "default", "prefer_more" or "prefer_less". For more information visit the DeepL API docs link below.
         "FORMALITY": "<Your DeepL formality preference here>",
     },
 }
 ```
+
+[DeepL API docs](https://www.deepl.com/docs-api/translate-text/)
 
 ## LibreTranslate
 
--- a/docs/how-to/integrations/machine-translation.md
+++ b/docs/how-to/integrations/machine-translation.md
@@ -124,6 +124,8 @@ WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
     "CLASS": "wagtail_localize.machine_translators.deepl.DeepLTranslator",
     "OPTIONS": {
         "AUTH_KEY": "<Your DeepL key here>",
+        # Optional DeepL API option
+        "FORMALITY": "<Your DeepL formality preference here>",
     },
 }
 ```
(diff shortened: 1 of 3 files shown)


Abdul-Dridi:
@zerolab Any chance I could request a review on this? 🙏

wagtail_localize/machine_translators/deepl.py
```diff
@@ -24,17 +24,20 @@ def get_api_endpoint(self):
         return "https://api.deepl.com/v2/translate"
 
     def translate(self, source_locale, target_locale, strings):
+        args = {
```
zerolab, line 27: nitpick: let's call this `parameters`

docs/how-to/integrations/machine-translation.md
```diff
@@ -124,6 +124,8 @@ WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
     "CLASS": "wagtail_localize.machine_translators.deepl.DeepLTranslator",
     "OPTIONS": {
         "AUTH_KEY": "<Your DeepL key here>",
+        # Optional DeepL API option
+        "FORMALITY": "<Your DeepL formality preference here>",
```
zerolab, line 128: Reading https://www.deepl.com/docs-api/translate-text we should
1. link to the docs
2. highlight that only "default", "prefer_less", "prefer_more" will work across all locales.
Even better, we should only accept these three

zerolab (changes requested):
Thanks @Abdul-Dridi 

Left a couple of notes. The main note is around the fact that the requests will fail for unsuported languages if set to "less" or "more"

docs/how-to/integrations/machine-translation.md
Abdul-Dridi: Sounds good!

How would you like me to handle the case in which someone provides an unsupported formality option?

- Raise a ValueError
- Ignore the provided option and don't set formality

I personally prefer the first

zerolab: I'd say raise a warning and ignore

Abdul-Dridi: Updated 👍

Abdul-Dridi:
Good for another review 👍

docs/how-to/integrations/machine-translation.md
```diff
@@ -124,10 +124,14 @@ WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
     "CLASS": "wagtail_localize.machine_translators.deepl.DeepLTranslator",
     "OPTIONS": {
         "AUTH_KEY": "<Your DeepL key here>",
+        # Optional DeepL API option. Accepts "default", "prefer_more" or "prefer_less". For more information visit the DeepL API docs link below.
+        "FORMALITY": "<Your DeepL formality preference here>",
     },
 }
 ```
 
+[DeepL API docs](https://www.deepl.com/docs-api/translate-text/)
```
zerolab, line 133: I think this should just go in the example code comment. Will tidy up when I come to merge

Abdul-Dridi: Sounds good, did it like this because hyperlinks don't format properly inside the python block

zerolab (approved):

zerolab:
Thanks. Tidied up and merged in cf12512

Issue #786, opened 2024-03-05
feature request: Machine translate and publish


onno-timmerman:
When you sync a page there is this nice option "immediate publish"
Could there also be an option "Machine translate and plublish" so the translation is already done by Deepl etc..

![image](https://github.com/wagtail/wagtail-localize/assets/178425/7fcaf95d-19c1-45f4-8b69-4e89b94daf81)

zerolab:
Ooh, I like this idea. We'd need:

1. to detect we have a machine translator configured
2. change the checkbox to a radio button or select so only one option is allowed. Alternatively, add some JS to untick the other option when one of them is ticked
3. trigger the machine translation workflow after sync, then do the publishing

waldo90:
I'm currently developing an automatic machine translation feature on a client's website - I feel it's pretty close to what's being discussed here. Currently the implementation is external to `wagtail-localize` and lacks the UI selection feature - it hooks the `after_create_page` event and machine translates to all configured locales.


Something like:

## wagtail_hooks.py
```python
...
@hooks.register('after_create_page')
def after_create_page(request, page):

    locales_translated = machine.create_machine_translations(page, request.user)
    if request.POST.get('action-publish') == 'action-publish':
        machine.publish_translations(page, request.user)
        machine.stop_translations(page, request.user)

    langs = [l.language_name for l in locales_translated]
    langs_string = " and ".join([", ".join(langs[:-1]),langs[-1]])
    message = f"{page.locale.language_name} page has also been machine translated into { langs_string }"

    messages.success(
        request,
        message,
    )
```

## machine.py

```python
def create_machine_translations(page, user):
    """
    Progress untranslated locale pages to fully translated pages
    under Translation workflow
    """
    locales = [locale for locale in Locale.objects.all() if locale != page.locale]

    translate_object( page, locales, user=user)

    translations = Translation.objects.filter(source__object_id=page.translation_key)
    translator = get_machine_translator()

    for translation in translations:
        # Get segments
        segments = defaultdict(list)
        for string_segment in translation.source.stringsegment_set.all().select_related(
            "context", "string"
        ):
            segment = StringSegmentValue(
                string_segment.context.path, string_segment.string.as_value()
            ).with_order(string_segment.order)
            if string_segment.attrs:
                segment.attrs = json.loads(string_segment.attrs)

            # Don't translate if there already is a translation
            if StringTranslation.objects.filter(
                translation_of_id=string_segment.string_id,
[...]

PR #807, opened 2024-06-19
Add a machine translation option when syncing pages

waldo90:
This PR adds an option for machine translation to be performed when synicing page translations as per issue:
[#786](https://github.com/wagtail/wagtail-localize/issues/786#issue-2169754390)

It adds the option alongside the existing 'publish immediately' option as I found that these can be made to work togther in all combinations:
 - sync only
 - sync and publish without machine translation
 - sync and publish with machine tranlation
 - sync with machine translation (unpublished but ready for review)

This is slightly as odds with what was discussed on the ticket. However, I'm hopeful that the extra flexibility will be welcome.

### DIFF
--- a/wagtail_localize/views/edit_translation.py
+++ b/wagtail_localize/views/edit_translation.py
@@ -1370,18 +1370,14 @@ def upload_pofile(request, translation_id):
     return redirect(next_url)
 
 
-def apply_machine_translation(translation_id, user):
+def apply_machine_translation(translation_id, user, machine_translator):
     translation = get_object_or_404(Translation, id=translation_id)
 
     instance = translation.get_target_instance()
     if not user_can_edit_instance(user, instance):
         raise PermissionDenied
 
-    translator = get_machine_translator()
-    if translator is None:
-        raise Http404
-
-    if not translator.can_translate(
+    if not machine_translator.can_translate(
         translation.source.locale, translation.target_locale
     ):
         raise Http404
@@ -1410,7 +1406,7 @@ def apply_machine_translation(translation_id, user):
         )
 
     if segments:
-        translations = translator.translate(
+        translations = machine_translator.translate(
             translation.source.locale, translation.target_locale, segments.keys()
         )
 
@@ -1424,7 +1420,7 @@ def apply_machine_translation(translation_id, user):
                         defaults={
                             "data": translations[string].data,
                             "translation_type": StringTranslation.TRANSLATION_TYPE_MACHINE,
-                            "tool_name": translator.display_name,
+                            "tool_name": machine_translator.display_name,
                             "last_translated_by": user,
                             "has_error": False,
                             "field_error": "",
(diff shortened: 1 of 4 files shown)


wagtail_localize/views/update_translations.py
```diff
@@ -17,16 +17,35 @@
...
+if get_machine_translator():
+    PUBLISH_TRANSLATIONS_HELP = (
+        "Select this to publish immediately. Changes will be published "
+        "in the original language unless you also select use machine translation."
+    )
+else:
+    PUBLISH_TRANSLATIONS_HELP = (
+        "Select this to publish immediately. Changes will be published "
+        "in the original language until you translate and publish them."
+    )
```
zerolab, line 35: I think these need tweaking as there is a discrepancy in style between this and the use machine translation help.

I suggest something along the lines of:

> Apply the updates and publish immediately. The changes will use the original language until translated unless you also select "Use machine translation".

and 

> Apply the updates and publish immediately. The changes will use the original language until translated.

wagtail_localize/views/edit_translation.py
```diff
@@ -1426,12 +1425,19 @@ def machine_translate(request, translation_id):
...
                             "field_error": "",
                         },
                     )
+        return True
+    return False
 
+
+@require_POST
+def machine_translate(request, translation_id):
+    if apply_machine_translation(translation_id, request.user):
```
zerolab, line 1439: `apply_machine_translation` is doing too many things and can have unintended consequences on the sync translation form.

I would rather it purely takes the translation id, user, and the machine translator instance and it *applies* the machine translation.

Move the additional checks in the view (and ditto, as additional checks on the sync translation form init and validation)

zerolab (changes requested):
This is a great start @waldo90 

Left a few notes and suggestions for improvement

wagtail_localize/views/update_translations.py
waldo90: Thank you, I've used your copy verbatim.

wagtail_localize/views/update_translations.py
```diff
@@ -17,16 +17,35 @@
...
 class UpdateTranslationsForm(forms.Form):
     publish_translations = forms.BooleanField(
         label=gettext_lazy("Publish immediately"),
-        help_text=gettext_lazy(
-            "This will apply the updates and publish immediately, before any new translations happen."
-        ),
+        help_text=gettext_lazy(PUBLISH_TRANSLATIONS_HELP),
+        required=False,
+    )
+    use_machine_translation = forms.BooleanField(
```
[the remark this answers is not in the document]
waldo90: The checkbox rendering is bypassed in the template `update_translations.html` around line 53 which checks if the machine translator is available. It's a different approach to what you are thinking but what do you think? Is that ok?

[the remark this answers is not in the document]
waldo90: Added tests foor the two form states

zerolab:
Thanks @waldo90. Will finish the work in #842 (mainly rebase and a bit of tidy up)

PR #819, opened 2024-08-27
Let machine translators skip fields

tognee:
Currently there is no way to keep a field not translated in case of issues with the machine translator.
Setting the value to an empty StringValue will show the field as translated in the admin.

This patch will let machine translators omit the key value pair to keep the field not translated.

(Sorry for the double PR, had issues with my linter)

### DIFF
--- a/wagtail_localize/tests/test_edit_translation.py
+++ b/wagtail_localize/tests/test_edit_translation.py
@@ -33,6 +33,7 @@ from wagtail.images.tests.utils import get_test_image_file
 from wagtail.models import Locale, Page, Revision
 from wagtail.test.utils import WagtailTestUtils
 
+from wagtail_localize.machine_translators.dummy import translate_html
 from wagtail_localize.models import (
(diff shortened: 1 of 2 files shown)


zerolab:
Any chance you could add a test or two?

tognee:
To add tests I would need to add a machine translator that could fail.
Can I edit the dummy translator so that it doesn't generate a response if the text to translate is more than 50 words, or should I add another machine translator just for the test?

zerolab:
You could [mock patch](https://docs.python.org/3/library/unittest.mock.html) https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/machine_translators/dummy.py#L23 for your test only

tognee:
For my first implementation I would need to patch the translate method in the DummyTranslator class. I've extended the implementation to check if the StringValue data is null.

Added tests for both of them and it seems to be working.

zerolab (approved):

PR #824, opened 2024-09-05
Add support for syncing translations from unpublished revisions

dases:
This is intended to provide for syncing Translation source changes that haven't yet been published.

Currently, once a `TranslationSource` is created it is frozen until the source is published. This can be a problem when a translation is created early in the content creation process: updates to the source cannot be synced until the source is published, which may not be desirable.  
Supporting pre-publish and ongoing draft workflows is essential, imo.

This PR adds another field to the Update Translations form which allows for updating the `TranslationSource` from the latest revision.

The code here is a first draft - a prompt for discussion - because workflow-wise there are things to consider:
- If the source for `TranslationSource` has *never* been published, perhaps we just *always* sync from the latest revision?
- It's potentially surprising when you've previously updated from latest revision and you forget to enable it on the "Update Translations" form (the `TranslationSource` reverts to what is current in source object, which may not be the latest revision.)
- Does Sync from draft option even belong in the "Update Translations" form - would it be better as another option in the header buttons? (assuming draft is draft, and we don't want to put this next to a "Publish immediately" option)

Will add tests if/when an acceptable approach to this is agreed on.

### DIFF
--- a/wagtail_localize/views/update_translations.py
+++ b/wagtail_localize/views/update_translations.py
@@ -13,7 +13,7 @@ from django.utils.translation import gettext_lazy
 from django.views.generic import TemplateView
 from django.views.generic.detail import SingleObjectMixin
 from wagtail.admin.views.pages.utils import get_valid_next_url_from_request
-from wagtail.models import Page
+from wagtail.models import Page, RevisionMixin
 from wagtail.snippets.models import get_snippet_models
 from wagtail.utils.version import get_main_version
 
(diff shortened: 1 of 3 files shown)


wagtail_localize/views/update_translations.py
```diff
@@ -29,6 +29,20 @@ class UpdateTranslationsForm(forms.Form):
         ),
         required=False,
     )
+    from_revision = forms.BooleanField(
+        label=gettext_lazy("Sync draft changes"),
+        help_text=gettext_lazy(
+            "This will update translations with current draft content. "
+            "If not set, then only published changes will be synced."
```
dases, line 36: micro-copy fail: this not quite true if the source has never been published.

dases:
The main issue this PR is addressing - not being able to sync translations from changes to a page if it has not been published - is not an issue on Wagtail > 6. as a result of: https://docs.wagtail.org/en/stable/releases/6.0.html#changed-handling-of-database-updates-of-non-live-page-objects-or-subclasses-of-draftstatemixin

PR #829, opened 2024-10-09
Fix getting progress when there are no strings

alexkiro:
Unlike COUNT, SUM will actually return NULL not 0 when there are no rows, so we need to coalesce to 0 to avoid such cases.

This issued caused both the progress and status values to be incorrect for translations. Since `0 != None` an empty page will have its translations marked as "waiting for translations."


zerolab:
Thank you @alexkiro

PR #830, opened 2024-10-09
Add support for filtering the translation report by status

alexkiro:
Adds a new filter for filtering by the status of the translation in the translation report.

Note that the filter will not be consistent with the column for translations with sources that don't have any strings, until #829 is merged. So it will appear to not function correctly for such cases.


zerolab (approved):
Nice one, thanks @alexkiro

PR #835, opened 2024-10-28
Add more options to the DeepL settings (Fixes #831)

### LINKED ISSUE #831: Allow the use of Glossaries when using DeepL
DeepL allows the usage of Glossaries when translating text so users can provide their own translations for certain text.
It would be great to allow specifying a glossary ID from within the settings.

https://developers.deepl.com/docs/api-reference/glossaries

alexkiro:
- Add the ability to configure glossaries by (source, target) language pairs.
- Add the ability to configure the timeout instead of having it hardcoded to 30.
- Add tests for the new glossary options.
- Update docs for the new options.

Fixes #831

---

I've also extracted constructing the parameters into a separate method. This way if anyone else needs even more advanced customizations of the parameters it's easier to extend:

```python
from wagtail_localize.machine_translators.deepl import DeepLTranslator


class CustomDeepLTranslator(DeepLTranslator):
    def get_parameters(self, source_locale, target_locale, strings):
        parameters = super().get_parameters(source_locale, target_locale, strings)
        parameters["another_option"] = "My custom value"
        return parameters
```

### DIFF
--- a/docs/how-to/integrations/machine-translation.md
+++ b/docs/how-to/integrations/machine-translation.md
@@ -127,6 +127,13 @@ WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
         # Optional DeepL API setting. Accepts "default", "prefer_more" or "prefer_less".\
         # For more information see the API docs https://www.deepl.com/docs-api/translate-text/
         "FORMALITY": "<Your DeepL formality preference here>",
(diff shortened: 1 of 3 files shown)


zerolab (approved):

zerolab:
Thanks @alexkiro

PR #836, opened 2024-10-29
Some small gotcha with snippets

onno-timmerman:
When using snippets and translations there is a gotcha you might encounter. This might save somebody couple of hours figuring it out.

Imagine you have a page with related tags added via snippets, and you want to query all pages with a specific tag.

When synchronous translation is enabled, the tag will refer to the default language. However, if the user turns off
synchronous translation, the tag will refer to the page’s language. This could mean the user references a translated tag
within the translated page, causing the query to miss some pages tagged with the original tag.

To solve this, you can query both the default language and the current language, then filter based on the active
language.

### DIFF
--- a/docs/how-to/tips.md
+++ b/docs/how-to/tips.md
@@ -9,11 +9,33 @@ To solve this, you can query both the default language and the current language,
 language.
 
 ```python
-def get_something(request, tag):
-  lang = request.LANGUAGE_CODE
-  default_tag = tag.get_translation(tag.get_default_locale())
-  pages = SomePage.objects.live().filter(
-    Q(locale__language_code=lang),
-    Q(tags=default_tag) | Q(tags=tag)
-  ).distinct()
+from django.db import models
+from django.db.models import Q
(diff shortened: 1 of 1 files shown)


docs/how-to/tips.md
```diff
@@ -0,0 +1,19 @@
+# Snippets and Synchronization
+Imagine you have a page with related tags added via snippets, and you want to query all pages with a specific tag.
```
zerolab, line 2: could you add a code snippet illustrating this?
it will help people contextualize the scenario

onno-timmerman: I'm not sure what you mean. I added a code snippet at the bottom ?

zerolab: the tags as snippets, and how they are added to the page

onno-timmerman: added full example, hope this helps

zerolab:
Thanks @onno-timmerman, merged in 2c2600a55f02f3f559fc124c9a334df3039743b7

PR #842, opened 2024-12-20
 Add a machine translation option when syncing pages (Rebase of #807)

zerolab:
This is a rebase and tidy up of #807


zerolab:
Thanks @waldo90

Issue #587, opened 2022-06-22
Video Files cannot be translated ('wagtailmedia.Media')


hennabanana:
Hi there,

currently its not possible to upload a Video for a Video-Media-File in a another language (https://github.com/torchbox/wagtailmedia). After Translating the Page, its just empty.

is it possible to add this datatype so that i can upload an video in another language by the MediaChooserPanel?

```
class VideoSlidePage(Page):

   video = models.ForeignKey(
        'wagtailmedia.Media',
        null=True,
        blank=True,
        verbose_name=_('Video'),
        on_delete=models.SET_NULL,
        related_name='+' )
```

![Screenshot 2022-06-22 at 08 32 07](https://user-images.githubusercontent.com/25147943/174960659-d41b0621-001b-41eb-bd49-96c724ce0a02.png)

zerolab:
Hey @hennabanana,
thank you for bringing this up.

Hmm, I would've thought this would be handled by the fallback logic. Will try to have a look sometime soon.

zerolab:
Had a look over the weekend. This will require more work as localize doesn't know which chooser to use.

catilgan-nextension:
Hi @zerolab 

Is there any chance that this feature will be implemented in near future? We need support for translating the audio files in StreamField fields. Currently, the UI in translation view only shows the foreign key of the media model.

I tried to implement the functionality by myself within the test application, but it's rather a "hacky" solution. So, we would be glad, if there is an official support for media files.

Thanks for the great work on the Wagtail Localize and Wagtail Media projects.

catilgan-nextension:
Hi @zerolab 

I implemented the missing chooser logic in our own fork and created a pull-request (#847). I hope my implementation helps to fix this issue. 

Unfortunately, the dependency checks are failing due to code formatting issues. Thanks for the support.

zerolab:
Thanks @catilgan-nextension, will try to have a look soon. The main consideration is we can't make `wagtailmedia` a dependency, so the code would need to check that `wagtailmedia` is an installed app

catilgan-nextension:
Thanks @zerolab for your reply and looking into this issue. 

Yes, i understand your point with the dependency of `wagtailmedia`, since it's not a Wagtail core module.  Nonetheless, the `wagtailmedia` is such an important module for most Wagtail sites, at least for us, the localization workflow should to be working for media files. This dependency might be "acceptable" in this case.

zerolab:
[...]

PR #847, opened 2025-01-15
Implemented and added media chooser into the edit translation workflow for audio and video files

catilgan-nextension:
Since the issue #587 still open, i tried to implement this missing feature. We needed this feature to translate the linked media files (Audio only) in our project as soon as possible.

Unfortunately, due to the missing deep knowledge in Wagtail module development process and time constraints, I've only implemented the necessary changes for this feature and skipped the unit tests. Furthermore, I couldn't test the video files, since we only need audio files in our project. I believe that the video files should work fine as well.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -82,6 +82,7 @@ basepython = python3.12
 deps =
     wagtail>=5.2,<6.3
     Django>=4.2,<5.2
+    wagtailmedia>=0.0.0,<2.0
 
 commands_pre =
     python {toxinidir}/testmanage.py makemigrations
--- a/wagtail_localize/templates/wagtail_localize/admin/edit_translation.html
+++ b/wagtail_localize/templates/wagtail_localize/admin/edit_translation.html
@@ -53,6 +53,7 @@
             window.chooserUrls = {
                 imageChooser: "{% url "wagtailimages_chooser:choose" %}",
                 documentChooser: "{% url "wagtaildocs_chooser:choose" %}",
+                mediaChooser: "{% url "wagtailmedia:chooser" %}",
                 pageChooser: "{% url "wagtailadmin_choose_page" %}",
             }
         }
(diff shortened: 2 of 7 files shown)


tox.ini
```diff
@@ -82,6 +82,7 @@ basepython = python3.12
 deps =
     wagtail>=5.2,<6.3
     Django>=4.2,<5.2
+    wagtailmedia>=0.0.0,<2.0
```
zerolab, line 85: needs to move to dev dependencies

wagtail_localize/templates/wagtail_localize/admin/edit_translation.html
```diff
@@ -53,6 +53,7 @@
             window.chooserUrls = {
                 imageChooser: "{% url "wagtailimages_chooser:choose" %}",
                 documentChooser: "{% url "wagtaildocs_chooser:choose" %}",
+                mediaChooser: "{% url "wagtailmedia:chooser" %}",
```
zerolab, line 56: To make this work without the hard dependency, we'll need to move the chooserUrls logic to a template tag and conditionally add `mediaChooser`.

wagtail_localize/views/edit_translation.py
```diff
@@ -366,6 +367,9 @@ def widget_from_block(block, content_components=None):
         elif isinstance(block, ImageChooserBlock):
             return {"type": "image_chooser"}
 
+        elif isinstance(block, AudioChooserBlock) or isinstance(block, VideoChooserBlock):
```
zerolab, line 370: wrap in an `if apps.is_installed("wagtail.embeds")` and move the block imports inline

https://github.com/wagtail/wagtail-localize/blob/b619a32cc6410fec28a27c0ce1e305605dcb44c0/wagtail_localize/segments/ingest.py#L119

wagtail_localize/views/edit_translation.py
```diff
@@ -366,6 +367,9 @@ def widget_from_block(block, content_components=None):
...
+            return {"type": "media_chooser"}
+
```
zerolab, line 372: also, need to check whether the direct field (that is the ForeignKey) is to a media class and set the widget type for it as well in `widget_from_field()`, a la https://github.com/wagtail/wagtail-localize/blob/b619a32cc6410fec28a27c0ce1e305605dcb44c0/wagtail_localize/views/edit_translation.py#L321-L322

zerolab (commented):
Thank you for starting on this @catilgan-nextension 
Had a quick look and left a few notes and suggestions.

This will need a few tests too.

wagtail_localize/templates/wagtail_localize/admin/edit_translation.html
catilgan-nextension: As suggested, I create a new template tag method which checks the wagtailmedia installation and adds the "wagtailmedia:chooser" conditionally into the translation templates. My local tests did work on test environment instance.

wagtail_localize/views/edit_translation.py
catilgan-nextension: I moved the blocks into an inline import if the wagtailmedia has been installed. The code structure didn't change.

PR #856, opened 2025-02-13
Fixed IndexError when a translation with overridden and unchanged Chooser inside a ListBlock is published

### LINKED ISSUE #855: ListBlock with PageChooserBlock can cause IndexError with mixed translations
Having a PageChooserBlock inside of a ListBlock allows Editors to leave some of the choosen pages untranslated and only changing the ones needed. Which is nice. But doing this causes an IndexError when publishing.

Example:
```python
class TestPage(Page):
    content = StreamField(StreamBlock([
        ('links', ListBlock(PageChooserBlock(), icon='link')),
    ]))
```

Now create a source language page with two references and translate into a target language. When only one of these pages is overridden by the translation, then wagtail crashes on publishing.

Error is thrown here in `handle_related_object`, because segments is empty for the not overridden page:
https://github.com/wagtail/wagtail-localize/blob/080064749f59883d06cb7d15d81ca095ed28c410/wagtail_localize/segments/ingest.py#L88

But think that `handle_list_block` should handle this case by not calling `self.handle_block` on empty segments:
https://github.com/wagtail/wagtail-localize/blob/080064749f59883d06cb7d15d81ca095ed28c410/wagtail_localize/segments/ingest.py#L206-L219

I implemented a basic hot-fix for this issue in my code:
```python
for block_index, block in enumerate(list_block.bound_blocks):
    block_segments = segments_by_block[block.id]
    if len(block_segments):
        list_block.bound_blocks[block_index].value = self.handle_block(
            block.block, block.value, block_segments
        )
```

That simple if seems to resolve this issue. Will test more and open a pr later.

murray3k:
This affected all `Chooser`s inside a `ListBlock` , when only one of them is translated and the other kept unchanged.
I added a simple empty check to `handle_list_block` so that `self.handle_block` is not called when there are no `block_segments` , which was causing the `IndexError` later. This fixes #855.

Added two Tests. One with images in a list where only one is changed and another one for a list with `PageChooserBlock`s. Might be redundant, but I wanted to test that this issue is with all choosers and not just the `PageChooserBlock` where we encountered it.

### DIFF
--- a/wagtail_localize/segments/ingest.py
+++ b/wagtail_localize/segments/ingest.py
@@ -212,9 +212,10 @@ class StreamFieldSegmentsWriter:
 
         for block_index, block in enumerate(list_block.bound_blocks):
             block_segments = segments_by_block[block.id]
-            list_block.bound_blocks[block_index].value = self.handle_block(
-                block.block, block.value, block_segments
-            )
+            if len(block_segments):
+                list_block.bound_blocks[block_index].value = self.handle_block(
+                    block.block, block.value, block_segments
+                )
 
         return list_block
 
--- a/wagtail_localize/segments/tests/test_segment_ingestion.py
+++ b/wagtail_localize/segments/tests/test_segment_ingestion.py
@@ -707,6 +707,134 @@ class TestSegmentIngestionWithStreamField(TestCase):
             ],
         )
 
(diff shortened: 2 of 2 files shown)


zerolab (approved):
Nice one. Thank you @murray3k

PR #859, opened 2025-03-03
Add support for RevisionMixin models that are not also DraftStateMixin subclasses

### LINKED ISSUE #857: Snippets with RevisionMixin but not also DraftStateMixin crash.
Hi there and thanks for this great library.

I would like to draw your attention to this code here:

https://github.com/wagtail/wagtail-localize/blob/7797bd3d49e5c49fdbffc15f715ad542ae895264/wagtail_localize/models.py#L807-L816

Taken from `TranslationSource.create_or_update_translation`.

Here a new translation is being saved. The code checks to see if the translated object is a `RevisionMixin` and if so it creates a new_revision and then publishes it.

`Revision` from `wagtail.models` defines a `publish()` method:

```python
    # wagtail/models/__init__.py
    def publish(
        self,
        user=None,
        changed=True,
        log_action=True,
        previous_revision=None,
        skip_permission_checks=False,
    ):
        return self.content_object.publish(
            self,
            user=user,
            changed=changed,
            log_action=log_action,
            previous_revision=previous_revision,
            skip_permission_checks=skip_permission_checks,
        )
```

Note that it in turn calls `self.content_object.publish`, where `self.content_object` is essentially the translated object we started with.

The problem is that adding `RevisionMixin` to, say, a snippet does not give it a `publish()` method, so you get an `AttributeError` on `self.content_object`.

`DraftStateMixin` does provide a `publish()` method so snippets that have both drafts and revisions are fine but those with just revisions crash.

Something along the lines of the following might fix this:

```python
elif isinstance(translation, RevisionMixin):
    # We copied another instance which may be live, so we make sure this one matches the desired state
     translation.live = publish
     translation.save()

    # Create a new revision of the Snippet
    new_revision = translation.save_revision(user=user)

     if publish and isinstance(translation, DraftStateMixin): # <--- extra check
         transaction.on_commit(new_revision.publish)
```

alexalligator:
Fixes #857

Notes:
- In order to trigger the transaction.on_commit hook in TranslationSource.create_or_update translation I have used a TransactionTestCase.


zerolab (approved):

zerolab:
LGTM. Thank you for this. Aiming to make a new release this week

PR #868, opened 2025-05-23
Fix French translations: Replace "Regions" with "Languages" for "Locale"

Luzzzi:
This PR updates the French translations by replacing all occurrences of the term **"Locale"** previously translated as **"régions"** with **"langues"**.

The term "locale" in a technical context primarily relates to language and linguistic preferences rather than geographic regions. Using "regions" was misleading, as a region does not necessarily correspond to a single language, which could cause confusion. For example, Switzerland is one region but has multiple languages. Translating "locale" as "regions" could confuse users, whereas "languages" better reflects the concept and is more intuitive for French-speaking users.

Replacing it with "languages" provides a clearer and more accurate translation that better reflects the concept of locale from the users’ perspective.

Thank you for considering this improvement to make the interface more intuitive for French-speaking users.


zerolab:
Hi, sorry for the delayed response. Thank you for your contribution.

We fetch the translations from Transifex - https://explore.transifex.com/torchbox/wagtail/ as per https://docs.wagtail.org/en/stable/advanced_topics/i18n.html#wagtail-admin-translations. Which means these changes will get overriden next time I fetch translations.

Unfortunately cannot accept the PR, sorry about that.

Issue #869, opened 2025-06-03
DeepL API key must be sent in the HTTP authorisation header


catilgan-nextension:
According to the "Breaking changes (Changes Notices)" of **March 2025** in the DeepL API documentation, the API key must be send in an HTTP authorisation header, instead of the POST body data. 

In current implementation of `wagtail_localize/machine_translators/deepl.py`, we receive a `403 - Forbidden` answer from the DeepL API.

>  You will no longer be able to authenticate a request to any endpoint by sending an API key in a query parameter. Instead, send your API key in an HTTP header named Authorization .

Refer to DeepL API Documentation for further explanation: https://developers.deepl.com/docs/resources/breaking-changes-change-notices/march-2025-deprecating-get-requests-to-translate-and-authenticating-with-auth_key#change-notice

zerolab:
Will make a release in the next few hours

PR #870, opened 2025-06-03
Sets the DeepL API key in the HTTP authorisation header

catilgan-nextension:
With this PR the DeepL API key is set within the HTTP authorisation headers, which could've been set in the request body in the earlier DeepL API versions. Since March 2025, due to security enforcements, this approach is forbidden.

This PR fixes the #869 .


kaij:
Would be great to see this merged, fixes DeepL integration in Wagtail...

zerolab (approved):

PR #883, opened 2025-10-18
Feature: When adding a new locale, allow users to set publishing status of sync'd content

ThinksFast:
Hi, I am submitting a PR which enables a new feature when adding a new locale to wagtail.

When syncing a content tree from another language, the current behavior of wagtail localize is to mirror both the content and the publishing status. There is no way to override the publishing status in the UI. If an article is live in the source language, a newly synced locale and content tree will be automatically pushed live, before translation can be done. This workflow also does not give users the opportunity to configure other parts of the site, like translated navigation menus.

The new behavior added by this PR allows the user to configure the publishing status of all pages I the synced content tree. Two options are available:

Mirror: this is the current behavior and is configured as the default setting for backwards compatibility
Draft: all pages in the synced tree will be set to Draft
The code for this PR was assisted by AI. I've tested the behavior on my local machine and it works as I expected. Tests are included as well. Happy to take any feedback and make changes as needed.

image

### DIFF
--- a/wagtail_localize/locales/tests.py
+++ b/wagtail_localize/locales/tests.py
@@ -130,7 +130,7 @@ class TestLocaleCreateView(BaseLocaleTestCase):
 
         response = self.post(post_data)
 
-        # Should redirect back to index
+        # Should redirect back to index page
         self.assert_redirect_to_index(response)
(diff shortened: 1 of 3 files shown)


zerolab:
Hi, thank you for this. Will do my best to get this reviewed in the coming few weeks

PR #884, opened 2025-10-20
UpdateTranslationsView: sync draft content if available

### LINKED ISSUE #823: Wagtail-localize only syncs to the translated pages, sometimes.
Hello, 
I’m using Wagtail localize and having an issue getting the translated pages (in French) to sync to changes in the original language (English). It synced once in the beginning, but after that, I am unable to get the synchronization to go into the French page. Is there a reason why the synchronization won’t work after the initial translation? I noticed the “Translate with Dummy translator” button is greyed out in the destination page, and none of the new text I entered in the English page appears in the destination (French) page, after I save the draft and hit the “sync translated pages” button. It seems the target page did not detect the changes in the original language, even after the "sync translated pages" was pressed. Any idea on how to debug this or what’s happening would be greatly appreciated.I’m  not sure if this is normal behavior.  Below are relevant sections of my settings (base.py)  file:
INSTALLED_APPS = [
   "home",
   "search",
   "blog",
   "wagtail_localize",
   "wagtail_localize.locales",
MIDDLEWARE = [
   "django.contrib.sessions.middleware.SessionMiddleware",
   "django.middleware.common.CommonMiddleware",
   "django.middleware.csrf.CsrfViewMiddleware",
   "django.contrib.auth.middleware.AuthenticationMiddleware",
   "django.contrib.messages.middleware.MessageMiddleware",
   "django.middleware.clickjacking.XFrameOptionsMiddleware",
   "django.middleware.security.SecurityMiddleware",
   "django.middleware.locale.LocaleMiddleware",
   "wagtail.contrib.redirects.middleware.RedirectMiddleware",
]
LANGUAGE_CODE = "en-us"
TIME_ZONE = "UTC"
USE_I18N = True
# pvwagtail: Enable Wagtail i18n
WAGTAIL_I18N_ENABLED = True
WAGTAIL_CONTENT_LANGUAGES = LANGUAGES = [
   ("en", "English"),
   ("fr", "French"),
]
WAGTAILLOCALIZE_MACHINE_TRANSLATOR = {
   "CLASS": "wagtail_localize.machine_translators.dummy.DummyTranslator",
}
 I'm a newbie with Wagtail... using Django 5.0.8, Python 3.11.9, Wagtail 6.2.1, Wagtail localize 1.9.1

th3hamm0r:
As noted in https://github.com/wagtail/wagtail-localize/issues/823#issuecomment-2479458882 the draft content is not synced, when the "Sync translated pages" action is used.
This allows to change/prepare the content of a source object, without publishing it (yet), and further sync those changes to the translations to prepare those changes to be published later all at once.

I've deviated slightly [from my suggestion to introduce a new setting](https://github.com/wagtail/wagtail-localize/issues/823#issuecomment-3421150938). Instead, the sync only uses the draft content if the "Publish immediately" checkbox is not checked.

What do you think about it? Better use a setting and disable that functionality by default?

Fixes #823

### DIFF
--- a/wagtail_localize/models.py
+++ b/wagtail_localize/models.py
@@ -451,15 +451,25 @@ class TranslationSource(models.Model):
         return source, created
 
     @transaction.atomic
-    def update_from_db(self):
+    def update_from_db(self, use_draft_content=False):
         """
         Retrieves the source instance from the database and updates this TranslationSource
         with its current contents.
+
+        Args:
+            use_draft_content: If True and the source instance is a subclass of DraftStateMixin, the content is
+                retrieved from either the currently scheduled revision, or from the latest revision.
 
         Raises:
             Model.DoesNotExist: If the source instance has been deleted.
         """
         instance = self.get_source_instance()
+
+        if use_draft_content and isinstance(instance, DraftStateMixin):
+            instance = (
+                instance.get_scheduled_revision_as_object()
+                or instance.get_latest_revision_as_object()
+            )
 
         if isinstance(instance, ClusterableModel):
             self.content_json = instance.to_json()
--- a/wagtail_localize/tests/test_update_translations.py
+++ b/wagtail_localize/tests/test_update_translations.py
@@ -729,3 +729,90 @@ class TestUpdateTranslations(TestCase, WagtailTestUtils):
             context_id=string_segment.context_id,
         )
         self.assertEqual(string_translation.data, "post blog Edited")
+
+    def test_post_update_draft_page_translation(self):
+        self.en_blog_post.test_charfield = "Edited blog post"
+        self.en_blog_post.save_revision()
+
+        response = self.client.post(
+            reverse(
+                "wagtail_localize:update_translations",
+                args=[self.page_source.id],
+            )
+        )
+
+        self.assertRedirects(
+            response, reverse("wagtailadmin_explore", args=[self.en_blog_index.id])
+        )
+
+        # Check that the new string was submitted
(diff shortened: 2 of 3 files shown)


lourduradjou:
I reviewed this code and It solves the issue #823, Attaching a video that showcases this.

https://github.com/user-attachments/assets/b7cd6236-c737-4885-adb9-b234da4cf767

###  My review on what is happening now:

**Two cases:**

**Case 1:** (Existing code without this PR changes applied)

- If I create a new page , draft save it, apply translate this page. I can see the translate view page.
- Here I can do some changes in the source page(assume english) and sync translate again that will reflect in the translated page (say french) so it works fine in this case.

**Case 2:** (Existing code without this PR changes applied)

- After the case one, I published the site.
- Now If I make changes and save draft it, apply sync translate. This time it won't work as now onwards it requires the changes to be published so that the sync translate would work. (This is the issue which performs opposite the case 1)

After applying author's PR changes locally in the code it handles both **CASE 1** and **CASE 2** (Tested by me locally). So the issue was addressed but It could be enhanced a little.

**Suggestions as a User:**

What I felt was each time I have to **publish/save draft** and then click the **three dots** to open the pop up to translate the page. This is very **redundant** in nature. Instead we can have a button say **Publish and sync** in the bottom view itself or **Save draft and sync** or **Publish and draft sync**. This approach would reduce the frequent need of opening the three dots. 

image

(Here we can add new buttons , maybe something which when hovered opens into more buttons "compactness :)" )


@zerolab if you prefer some changes in this PR or my suggestions I am more than happy to work in it and add @th3hamm0r as co-author to solve this issue.

Issue #885, opened 2025-10-20
UI: Page footer menu looks broken


th3hamm0r:
### Issue Summary

The action menu in the footer of the page edit screen looks a bit broken (but it works):

<img width="1131" height="725" alt="Image" src="https://github.com/user-attachments/assets/f9e73fe8-84fa-4bbc-80ce-361d6d5be0a6" />

### Steps to Reproduce

1. Install [wagtail bakery](https://github.com/wagtail/bakerydemo)
2. Install [wagtail-localize](https://github.com/wagtail/wagtail-localize/?tab=readme-ov-file#installation)
3. Go to Homepage and create a translation into another language (e.g., http://127.0.0.1:8000/admin/localize/submit/page/60/)
4. Open the translated page (e.g., http://127.0.0.1:8000/admin/pages/84/edit/)
5. Open the action menu in the footer

### Technical details

-   Django version: 5.2
-   Wagtail version: 7.0, 7.1
-   Browser version: Chrome 141, Firefox 143

PR #886, opened 2025-10-21
Fix some UI issues in the footer's action menu

th3hamm0r:
- This mostly replaces forms with w-action controllers.
- The convert-to-alias button used a form unnecessarily (it's just a link to a separate page).
- I've tested it with 6.3, 7.0 and 7.1 (bakerydemo)

**TODO:** Unfortunately the `w-action` controller does not support an "action" form field and therefore the publish-form is still required. So this minor UI issue remains as seen below (without the form the button container would be have the same width as the dropdown). It probably requires css fixes, where I'm not an expert in.
Bildschirmfoto vom 2025-10-21 10-30-55

Relates to #885


zerolab:
question: we still support Wagtail 6.3 (as LTS). Do the changes work there?

Issue #789, opened 2024-03-10
feature request: sync


onno-timmerman:
- Bulk sync.
- publish and sync

Imagine you adapt a bunch of articles 20 by cleaning up SEO and some titles. Then you need to sync one by one all these articles. That is a cumbursome task. 

Would it be possbile to have a option besides publish that would be publish and sync translations

As bonus it would be ideal if the sync translations also do the machine translation in one time. It would be a huge timesaver

zerolab:
Hey @onno-timmerman,

This is a sensible suggestion. I see two parts to this:

1. Register a page action via [`register_page_action_menu_item`](https://docs.wagtail.org/en/stable/reference/hooks.html#register-page-action-menu-item) that checks that the user can publish and that they can sync. And a snippet action via [`construct_snippet_action_menu`](https://docs.wagtail.org/en/stable/reference/hooks.html#construct-snippet-action-menu)
2. Register a bulk action via [`register_bulk_action`](https://docs.wagtail.org/en/stable/reference/hooks.html#register-bulk-action) for bulk actions

For machine translation: #786 would take care of part of the flow, so we would need to hook build on that.

th3hamm0r:
I wanna give this a try.

Since the sync feature causes the most problems for our customers, among other things because it's difficult to explain and easy to forget, and is often the cause of some kind of server errors, I'd like to add an auto-sync setting. I would leave it disabled by default (keeping the old behavior), meaning additional buttons would be added as discussed above.
However, if enabled, publishing would automatically trigger a sync. Additionally, there would be a button for "publish + machine-translate", if available.
So when enabled, the normal save action still won't sync automatically, but the publish action would.
What do you think of that @zerolab?

I'm not 100% sure if skipping the auto-sync for save actions makes sense to the user, but I also think that a user might try things out while editing the draft and so we should not sync that automatically. With that, the user still needs to know, that he has to sync manually to get his changes into the other languages.

A bit related: https://github.com/wagtail/wagtail-localize/discussions/703

th3hamm0r:
> I'm not 100% sure if skipping the auto-sync for save actions makes sense to the user, but I also think that a user might try things out while editing the draft and so we should not sync that automatically. With that, the user still needs to know, that he has to sync manually to get his changes into the other languages.

[...]

PR #887, opened 2025-10-21
Auto sync and machine translation

th3hamm0r:
This only partially implements the proposed changes of https://github.com/wagtail/wagtail-localize/issues/789#issuecomment-3414169506 and should be seen as an early draft now open for discussion.

**TBD:**
* Publication + automatic machine translation: the attribute on the request looks a bit ugly, but it was the easiest and in my view the lightest approach (in contrast to sub-classing the `EditView` for `Page`).
* I've left out snippets for now, but I could mostly mirror the page logic, but listen for the `wagtail.signals.published` signal instead of the `after_publish_page` hook (actually this signal could be used for pages too :thinking: )
* I've skipped bulk sync for now, as it requires quite a lot of extra code and UI
* The go_live_at date isn't synced at the moment (which is understandable, as there are other use cases too), so I had to skip the publish+machine-translation for future dates. I've ignored this in the auto-sync on publish action, as it looked sensible to me to get the new content into the translations when a publication is planned.

**Todo:**
- [ ] Tests
- [ ] Docs

Relates to #789

### DIFF
--- a/wagtail_localize/views/update_translations.py
+++ b/wagtail_localize/views/update_translations.py
@@ -23,6 +23,32 @@ from wagtail_localize.views.edit_translation import apply_machine_translation
 from wagtail_localize.views.submit_translations import TranslationComponentManager
 
 
+def update_translations(
+    request,
+    instance,
+    translations,
+    use_machine_translation=False,
+    publish_translations=False,
+):
+    instance.update_from_db()
+
+    if use_machine_translation:
+        machine_translator = get_machine_translator()
+        for translation in translations.select_related("target_locale"):
+            apply_machine_translation(translation.id, request.user, machine_translator)
+
+    if publish_translations:
+        for translation in translations.select_related("target_locale"):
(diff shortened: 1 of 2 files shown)


wagtail_localize/views/update_translations.py
```diff
@@ -153,28 +179,17 @@ def post(self, request, **kwargs):
...
+            self.request,
+            self.object,
+            enabled_translations,
+            use_machine_translation=form.cleaned_data.get(
+                "use_machine_translation", False
+            ),
+            publish_translations=form.cleaned_data["publish_translations"],
+        )
 
         self.components.save(
```
th3hamm0r: With publish+auto-machine-translate we cannot really handle those components, I think. With bulk actions we may add those components to the confirmation view. That's why I have not extracted them into `update_translations()`.

I also don't fully understand them. Are those components meant to be created once per translation request or once per translation source? I know, it's up to the dev, but it would be nice to get some infos on how this has been used so far.

PR #891, opened 2025-11-26
Added support for SynchronizedField on ManyToManyFields

tognee:
This PR lets users set a ManyToManyField as a SynchronizedField without recieving the error:
> Direct assignment to the forward side of a many-to-many set is prohibited. Use model_name.set() instead.

This does not solve the issue of whether a ManyToMany field should be translatable or syncronized, but gives the option for the user to syncronize it without errors.

This PR adds:
 - the code changes to handle many to many fields
 - a snippet model to test ManyToMany fields (with its migration)
 - 3 unit tests to check if the SynchronizedField is respected correctly for ManyToMany

### DIFF
--- a/wagtail_localize/fields.py
+++ b/wagtail_localize/fields.py
@@ -292,6 +292,11 @@ def copy_synchronised_fields(source, target):
                     if values:
                         setattr(target, field.attname, values)
 
+            elif isinstance(field, models.ManyToManyField):
+                source_manager = getattr(source, field.name)
+                target_manager = getattr(target, field.name)
+                target_manager.set(source_manager.all())
+
             else:
                 # For all other fields, just set the attribute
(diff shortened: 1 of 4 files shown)


tognee:
I've seen that bitnami moved all free postgresql images to bitnamilegacy. I changed it just to test the checks, but I think it could it be replaced with the official docker postgresql image

Marvinrose (commented):
The codecov report flags one missing line in test/models.py, worth checking which line isn't covered and adding a test case for it.
The bitnamilegacy image change in test.yml makes sense as a short-term fix, but as you noted it might be worth replacing with the official postgres Docker image for long-term stability.

tognee:
@Marvinrose I've aligned to main and added a test for the `__str__` function in the test model
Now Codecov seems happy

PR #894, opened 2025-12-17
feat: support for django6.0

Akay7:
* at dependency allowed to install Django 6
* tox updated to run tests across supported versions of packages


zerolab (approved):
Nice!
Suggested a small tweak, which I can commmit.

mhindery:
Thanks for merging this, could there also be a subsequent new release of wagtail-localize?

PR #901, opened 2026-02-24
Fix whitespace preservation 8895408090764671849

Azadron228:


wagtail_localize/strings.py
```diff
@@ -9,6 +13,19 @@
 INLINE_TAGS = ["a", "abbr", "acronym", "b", "code", "em", "i", "strong", "br"]
 
 
+class PreserveWhitespaceTags:
```
Marvinrose, line 16: This class returns True for every tag, meaning BeautifulSoup will now preserve whitespace inside all HTML elements including block-level ones like <div> and <p>. The existing INLINE_TAGS list suggests the codebase already distinguishes inline from block elements. Was it intentional to override this for all tags, or should PreserveWhitespaceTags only apply to inline tags?

PR #903, opened 2026-02-26
Fixes #851 - Missing pagination on locale index view

### LINKED ISSUE #851: It shows only 20 locales without any pager in Settings/Locales
I have enabled 22 locales, but when I am in Settings / Locales page (/admin/locales/) there is shown only 20 locales (and no pager to go to next page).

It should either show all locales at once (not limiting them to 20) or it should show pager to go to next page to see rest of them.

Marvinrose:
### What
Fixes #851 - The Settings, Locales page was capped at 20 locales with no way to access the rest.

### Why
`IndexView` used a custom template (`wagtaillocales/index.html`) that overrode Wagtail's generic `{% block listing %}` with its own table loop. This bypassed Wagtail's built-in pagination entirely, meaning sites with 21+ locales could not access or manage the hidden ones.

### Fix
- Removed `template_name` from `IndexView` so it uses Wagtail's generic index template, which includes pagination automatically
- Updated the custom template to extend the generic template while preserving the Usage column (page counts per locale)
- Updated `test_simple` to assert the generic template is used instead of the custom one

### Testing
All `TestLocaleIndexView` tests pass:

TestLocaleIndexView

### DIFF
--- a/wagtail_localize/locales/templates/wagtaillocales/index.html
+++ b/wagtail_localize/locales/templates/wagtaillocales/index.html
@@ -44,4 +44,8 @@
 
         </div>
     </div>
+    {% if is_paginated %}
+    {% include "wagtailadmin/shared/pagination_nav.html" with page_obj=page_obj %}
+    {% endif %}
+
 {% endblock %}
--- a/wagtail_localize/locales/tests.py
+++ b/wagtail_localize/locales/tests.py
@@ -98,29 +98,77 @@ class TestLocaleIndexView(BaseLocaleTestCase):
         self.assertIn(locale, response.context["locales"])
 
     @override_settings(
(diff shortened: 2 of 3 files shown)


Marvinrose:
I also noticed the 10 CI failures are caused by a pre-existing bug where Wagtail renamed produced_error_message to get_error_message in a newer version. The fix is to add produced_error_message = None to both CreateView and EditView so they're compatible with the latest Wagtail. I've opened issue #904 and submitted a fix in PR #905.

zerolab:
#905 is now merged. Can you rebase, and also add a test please?

Marvinrose:
Hi @zerolab 

Rebased and added pagination test as requested.

PR #907, opened 2026-03-03
Fixes #669: Log warning when ValidationError occurs during translation.

### LINKED ISSUE #669: Wrong error shown when trying to translate an object when a validation error occurs
First off, props for this great piece of software :pray:.

Imagine I want to translate a `Tag`, that inherits from `TagBase` (and `TranslatableMixin`). It will thus have a `slug` field with a `unique=True` attribute.
When translating from the admin, the `TranslationSource` is created fine, but the target instance, another instance of `Tag` but with another `locale`, fails to be created because the database prevents saving another object with the same `slug`.

Because of this line of code : https://github.com/wagtail/wagtail-localize/blob/v1.3.3/wagtail_localize/operations.py#L81, the `ValidationError` is simply `pass`ed silently, and the error is hard to debug.
The error shown is `Tag matching query does not exist` that occurs later because the translated object does not exist. Full [traceback here](https://pastebin.com/rUmZvkNW).

Suggestions : 

- remove the `try / except ValidationError` clause and let the user see the `ValidationError`. I am not so familiar with the internals of wagtail-localize, but I guess there is a good reason for this try/except to be here
- add a warning to show the `ValidationError`

If it seems relevant for you, I'd be happy to try a PR to add the warning.

Marvinrose:
Previously, when `save_target()` raised a `ValidationError` during translation, the error was silently suppressed. This caused a confusing  `DoesNotExist` error to surface to the user instead of the real cause.

### Changes
- `operations.py`: replaced `contextlib.suppress(ValidationError)` with a  `logger.warning()` so the actual error is visible in logs.
- `views/submit_translations.py`: wrapped `get_translation()` with  `contextlib.suppress(DoesNotExist)` so the view degrades gracefully when `save_target()` fails
- Added a test to verify the warning is logged

### DIFF
--- a/wagtail_localize/views/submit_translations.py
+++ b/wagtail_localize/views/submit_translations.py
@@ -1,3 +1,5 @@
+import contextlib
+
 from django import forms
 from django.contrib import messages
 from django.contrib.admin.utils import quote, unquote
@@ -147,9 +149,10 @@ class SubmitTranslationView(SingleObjectMixin, TemplateView):
         single_translated_object = None
         if len(form.cleaned_data["locales"]) == 1:
             locales = form.cleaned_data["locales"][0].get_display_name()
-            single_translated_object = self.object.get_translation(
-                form.cleaned_data["locales"][0]
-            )
+            with contextlib.suppress(self.object.__class__.DoesNotExist):
+                single_translated_object = self.object.get_translation(
+                    form.cleaned_data["locales"][0]
+                )
 
         else:
             # Note: always plural
--- a/wagtail_localize/views/submit_translations.py
+++ b/wagtail_localize/views/submit_translations.py
@@ -1,5 +1,3 @@
-import contextlib
-
 from django import forms
 from django.contrib import messages
 from django.contrib.admin.utils import quote, unquote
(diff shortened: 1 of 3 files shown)


wagtail_localize/views/submit_translations.py
```diff
@@ -147,9 +149,10 @@ def form_valid(self, form):
         single_translated_object = None
         if len(form.cleaned_data["locales"]) == 1:
             locales = form.cleaned_data["locales"][0].get_display_name()
-            single_translated_object = self.object.get_translation(
-                form.cleaned_data["locales"][0]
-            )
+            with contextlib.suppress(self.object.__class__.DoesNotExist):
```
zerolab: https://docs.astral.sh/ruff/rules/suppressible-exception/

> Note that `contextlib.suppress` is slower than using `try-except-pass` directly. For performance-critical code, consider retaining the `try-except-pass` pattern.

While this is not "performance critical", everything adds up

Marvinrose: Hi @zerolab 

Great point, reverted to try-except-pass. Thank you so much for this!

PR #910, opened 2026-03-10
Add publish-and-sync translations actions to editor and bulk workflows (Fixes #789)

### LINKED ISSUE #789: feature request: sync

zanzo2003:
Solved Issue - #789 
Summary
- introduce a shared sync_translation_source() helper and result dataclass so page/snippet updates, hooks, and actions reuse the same publish + machine-translation pipeline
- add a “Publish & sync translations” editor action that appears when translated copies exist, reusing Wagtail’s long-running button UX and posting the special submit value
- mplement a bulk “Publish & sync translations” action with confirmation template, permission checks, and descendant handling so editors can publish multiple pages and synchronize locales in one step
- expand regression coverage for the new button visibility, hook messaging/permissions, and bulk action behavior

### DIFF
--- a/wagtail_localize/bulk_actions.py
+++ b/wagtail_localize/bulk_actions.py
@@ -1,0 +1,56 @@
+from django.utils.translation import gettext_lazy as _
+from wagtail.admin.views.pages.bulk_actions.page_bulk_action import PageBulkAction
+
+from wagtail_localize.models import TranslationSource
+from wagtail_localize.views.update_translations import sync_translation_source
+
+
+class PublishAndSyncTranslationsBulkAction(PageBulkAction):
+    display_name = _("Publish & sync translations")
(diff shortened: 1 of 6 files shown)


zanzo2003:
PFA Attaching test result

Screenshot 2026-03-10 at 1 30 04 AM
Screenshot 2026-03-10 at 1 30 45 AM
Screenshot 2026-03-10 at 1 31 15 AM
Screenshot 2026-03-10 at 1 32 38 AM

Issue #502, opened 2022-01-07
Showing text diff and old translation (tagged as needs revision) for changes made in synchronized translation


jnagler-git:
The current behaviour: when changing a field e.g., a paragraph, in the default language the old translation of that field gets deleted after the synchronization and the new text does not show any hints of changes that have been made. Thus, a new translation is needed.

It would be nice to have a diff feature, that does not delete the old translation, but to show it with a text diff of the default language. Sometimes just a few words or single sentences change in paragraph, so the translator can save a lot of time with this feature.

To not just give a request, I was looking a bit around and found out that this might be done using Google's [diff_match_patch](https://pypi.org/project/diff-match-patch/) library in the following way:

```python
import diff_match_patch
dmp = diff_match_patch.diff_match_patch()
diffs = dmp.diff_main('This is my original paragraph.', 'This is my revised paragraph with new words.')
 # semantic cleanup seems important 
dmp.diff_cleanupSemantic(diffs)
print(diffs)  
```
Output: `[(0, 'This is my '), (-1, 'original'), (1, 'revised'), (0, ' paragraph'), (1, ' with new words'), (0, '.')]`

This could then be canonically written as "This is my ~~original~~ **revised** paragraph **with new words**." Having the old translation still available then makes it a lot easier to revise the changes instead of translating it from scratch.

Would be wonderful to have this feature considered for wagtail-localize.

saevarom:
I've had this issue too, and it is quite annoying when e.g. someone just makes a linebreak in the original text, the original text block becomes two entirely new text blocks with their own hashes of the contents. The translation for the original text is lost to the end user, even though it still exists in the database. To confirm that, you can download a .po file to see that the original text exists in the .po file, but marked as an obsolete message.

I would propose giving the end user access to "obsolete" messages so that you would only be a couple of clicks away from the original translation to restore it. 

I have a client with almost 400 pages, all of them translated to another language, many of them with a lot of text. It is quite a pain to lose translations just because someone introduced a linebreak in the original text!

Any thoughts here @zerolab ?

th3hamm0r:
Fyi: there is also a related discussion https://github.com/wagtail/wagtail-localize/discussions/394
It's really a pain, always having to backup the original translated content. It is very easy to loose content, even if it it is probably still in the database somewhere...

zerolab:
I agree this is annoying. Unfortunately I do not have much capacity in the next month or so. Any PRs fixing this are most welcome ;)

[...]

PR #912, opened 2026-03-19
feat: surface previous translation memory hints in editor ( Issue #502 )

zanzo2003:
## Summary ( PR for Issue #502 )
  - preload prior StringTranslations per segment so the React editor can reference them when the source text changes
  - expose historic translation/comment/author to the UI and render a “Previous translation” banner with a reuse shortcut
  - add a regression test that covers the stale-segment scenario for pages 

  ## Testing
  - python3 -m pytest wagtail_localize/tests/
  test_edit_translation.py::TestGetEditTranslationView::test_segments_include_previous_translation_when_source_changes
  --ds=wagtail_localize.test.settings
  - npm run build (after npm install)
  - Manual: change EN source text, open FR translation editor, verify banner + “Reuse previous translation” button saves
  cleanly
  
  
 ## Adding test and Browser screen shots
 
 
Screenshot 2026-03-19 at 9 08 50 PM

Screenshot 2026-03-19 at 9 13 50 PM

Screenshot 2026-03-19 at 9 08 00 PM

Screenshot 2026-03-19 at 9 18 50 PM

### DIFF
--- a/wagtail_localize/static_src/editor/components/TranslationEditor/segments.tsx
+++ b/wagtail_localize/static_src/editor/components/TranslationEditor/segments.tsx
@@ -269,7 +269,6 @@ const BlockSegments = styled.ul`
     background-color: var(--w-color-surface-header, var(--w-color-grey-50));
     padding: 0;
     margin: 0;
-
     > li {
         &.errored {
             background-color: var(--w-color-critical-50, #fee7e8);
@@ -278,21 +277,17 @@ const BlockSegments = styled.ul`
             border: 1px solid var(--w-color-critical-100) !important;
             border-left-width: 5px !important;
         }
-
         &.incomplete {
             // !important required to override the border-bottom rule just below
             border-left: 5px solid var(--w-color-warning-100) !important;
         }
-
         &.complete {
             // !important required to override the border-bottom rule just below
             border-left: 5px solid var(--w-color-positive-100) !important;
         }
-
         &:not(:last-child) {
             border-bottom: 1px solid var(--w-color-border-furniture, #eeeeee);
         }
-
         &:after {
             content: '';
             display: table;
@@ -317,7 +312,6 @@ const SegmentFieldLabel = styled.h4`
 const SegmentSource = styled.p`
     padding: 15px 20px;
     font-style: italic;
-
     &.title {
         font-size: 1.875rem;
         font-weight: 800;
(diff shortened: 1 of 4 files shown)


zanzo2003:
## Previous source and tests screenshots 

Screenshot 2026-03-19 at 8 21 26 PM

Screenshot 2026-03-19 at 8 55 12 PM

shreyasbhaskar2001:
## Some minor feedbacks!!
 
 - wagtail_localize/locales/templates/wagtaillocales/edit.html:29, wagtail_localize/templates/wagtail_localize/admin/
    _components.html:7, wagtail_localize/templates/wagtail_localize/admin/submit_translation.html:24, wagtail_localize/
    templates/wagtail_localize/admin/update_translations.html:36: the includes for wagtailadmin/shared/
    non_field_errors.html have dropped out, so form-level validation errors (including component-level ones) will no
    longer be rendered. This regresses fixes merged recently—please restore the includes (a rebase onto current main
    should bring them back).
  - wagtail_localize/locales/views.py:75 and wagtail_localize/locales/templates/wagtaillocales/index.html:3: removing
    paginate_by = 20 and the pagination block undoes the new locale list pagination. On sites with many locales this will
    make the index slow and unwieldy, and the corresponding tests were also deleted. Reinstating the pagination logic
    (again, via rebase) is important to avoid that regression.

  ## Highlights

  - Nice work enriching TranslationEditor with a previousTranslation payload and UI so translators can re-use the last
    human translation when the source string changes. This was a major pain point for the user, saw a discussion dating back to Jan 22
  - The backend changes in edit_translation.py efficiently gather the historical translation metadata and you added
    test_segments_include_previous_translation_when_source_changes, which gives good coverage of the new behaviour.
    
    
    Additionally @thibaudcolas , can you also check if these feedbacks are valid from you p.o.v ? I think this adds some good value to the product .

wagtail_localize/static_src/editor/components/TranslationEditor/segments.tsx
```diff
@@ -365,24 +399,19 @@ const SegmentToolbar = styled.ul`
...
     }
-
     .icon {
         width: 1.3em;
         height: 1.3em;
         vertical-align: text-bottom;
         margin-left: 10px;
-
         &--green {
             color: #15704d;
```
Phinart98, line 414: Looks like the closing `}` for `&--green` got removed here along with the blank lines. Without it, `&--red` ends up nested inside `&--green`, so it'd compile to `.icon--green--red` instead of `.icon--red` — which would break the red error icon styling.

zanzo2003: Hey @Phinart98 , thanks for pointing that out !! totally missed that brace. I have updated the PR. Please let me know if there is anything else ? Thanks!

Yemeni:
I am approaching it differently. I am using an action/trigger that automatically lists the fuzzy search from the segments memory of wagtail-localize and lists them and injects them (using stimulus js )

image

image

PR #914, opened 2026-03-28
Fix inconsistent position of translation buttons across admin contexts

### LINKED ISSUE #911: "Translate this page" button position is inconsistent across page menus
## The problem

The three-dot action menu for pages appears in three different contexts in the Wagtail admin. The order of actions is inconsistent across these contexts, causing the "Translate this page" option to appear in a different position depending on where the user accesses it.

### Position of "Translate this page"

| View | URL | Position |
|------|-----|----------|
| Page listing | `/admin/pages/` | 10 (last, requires scrolling) |
| Page index | `/admin/pages/<id>/` | 5 |
| Page editor | `/admin/pages/<id>/edit/` | 6 |

### Full order across all three menus

| # | Listing `/admin/pages/` | pri | Editor `/admin/pages/<id>/edit/` | pri | Index `/admin/pages/<id>/` | pri |
|---|------------------------|-----|----------------------------------|-----|---------------------------|-----|
| 1 | Edit | 2 | Edit | 10 | Edit | 10 |
| 2 | View live | 6 | Add child page | 15 | Move | 20 |
| 3 | Add child page | 8 | Move | 20 | Copy | 30 |
| 4 | Move | 10 | Copy | 30 | Delete | 50 |
| 5 | Copy | 20 | Delete | 50 | **Translate this page** | **60** |
| 6 | Delete | 30 | **Translate this page** | **60** | Unpublish | 60 |
| 7 | Unpublish | 40 | Unpublish | 60 | History | 65 |
| 8 | History | 50 | History | 65 | Sort menu order | 70 |
| 9 | Sort menu order | 60 | Sort menu order | 70 | | |
| 10 | **Translate this page** | **60** | | | | |

_`pri` = the `priority` value assigned to the button when registered via the Wagtail hook. Lower values appear first._

## Root cause

The buttons in these menus are added via two separate Wagtail extension hooks. wagtail-localize registers the "Translate this page" button in both:

- `register_page_listing_more_buttons` — called when rendering the **page listing** menu (`/admin/pages/`)
- `register_page_header_buttons` — called when rendering the **page index** (`/admin/pages/<id>/`) and **page editor** (`/admin/pages/<id>/edit/`) menus, distinguished by the `view_name` parameter (`"index"` or `"edit"`)

wagtail-localize registers the "Translate this page" button with `priority=60` in both hooks from the same function (`wagtail_hooks.py` lines 169 and 181). However, each hook defines its own priority values independently with no shared or documented scale:

- In `register_page_listing_more_buttons` values range from **2 to 60** — so `priority=60` is the maximum
- In `register_page_header_buttons` values range from **10 to 70** — so `priority=60` falls in the middle

When two buttons share the same priority, Wagtail sorts them **alphabetically by label**. This determines the final behavior:

- In the listing: `"Sort menu order"` (S) < `"Translate this page"` (T) → Translate ends up last
- In the editor/index: `"Translate this page"` (T) < `"Unpublish"` (U) → Translate appears before Unpublish

## Steps to reproduce

**1. Page listing** — go to `/admin/pages/` and open the three-dot menu next to any page. "Translate this page" appears last and requires scrolling to find.

Image

**2. Page index** — go to `/admin/pages/<id>/` and open the three-dot menu. "Translate this page" appears in position 5.

Image

**3. Page editor** — go to `/admin/pages/<id>/edit/` and open the three-dot menu. "Translate this page" appears in position 6.

Image

## Expected behavior

Actions that appear across all three menus should maintain a consistent relative order. In particular, "Translate this page" should appear in a similar position in all three contexts, making it easier to discover and use. A user who learns where to find the option in one context should be able to find it in the others without confusion.

### Proposed solution

**Short term:** Review the `priority` value assigned to the "Translate this page" button in `wagtail_hooks.py` (lines 169 and 181) so that it appears in a consistent and visible position across all three menus.

**Longer term:** It may also be worth opening an issue in the Wagtail core repository, since the priority scales of both hooks are independent and a shared convention could help plugin authors position their buttons more predictably.

### Alternative solutions

The most straightforward fix would be to change `priority=60` to a lower value in `wagtail_hooks.py` (lines 169 and 181). However, since both hooks share the same function and use different priority scales, any single value will produce a different relative position in each context. A more complete solution would require changes to how the button is registered across both hooks.

### Additional context

_No response_

### Working on this

Contribute to the discussion to help this move forward. Explain your use cases, discuss possible solutions, research how to adapt existing code, or compatibility concerns.

Once others agree with the proposed changes, anyone can contribute to this. View our [contributing guidelines](https://docs.wagtail.org/en/latest/contributing/index.html). Add a comment to the issue to explain what approach you are taking.

rypptc:
Fixes #911

### Description

The "Translate this page" button appears in different contexts within the Wagtail admin: Page Listing, Page Index, and Page Editor. The "Sync translated pages" button is also affected, as it shares the same priority value within each context.

Its position was inconsistent across these contexts because the hooks defining this behavior used the same priority value, even though each hook operates within a different priority range.

Beyond aesthetics and consistency, the main issue was that in the Page Listing view the button was not visible without scrolling.

As an immediate fix, an internal function has been introduced. This function receives the priority value as a parameter and assigns it through context-specific hook handlers.

### AI usage

Used AI assistance to understand the hook priority system and identify appropriate priority values for each context.

### DIFF
--- a/wagtail_localize/tests/test_submit_translations.py
+++ b/wagtail_localize/tests/test_submit_translations.py
@@ -19,6 +19,10 @@ from wagtail_localize.test.models import (
     TestWithTranslationModeDisabledPage,
     TestWithTranslationModeEnabledPage,
 )
+from wagtail_localize.wagtail_hooks import (
+    page_header_buttons,
+    page_listing_more_buttons,
+)
 
 from .utils import assert_permission_denied, make_test_page
 
@@ -992,3 +996,28 @@ class TestSubmitSnippetTranslation(WagtailTestUtils, TestCase):
         )
 
         assert_permission_denied(self, response)
+
+
+class TestTranslationButtonPriority(WagtailTestUtils, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.en_locale = Locale.objects.get()
+        cls.fr_locale = Locale.objects.create(language_code="fr")
+        cls.de_locale = Locale.objects.create(language_code="de")
+
+        cls.en_homepage = Page.objects.get(depth=2)
+
+        cls.en_blog_index = make_test_page(cls.en_homepage, title="Blog", slug="blog")
+
+    def setUp(self):
+        self.user = self.login()
+
+    def test_listing_buttons_priority(self):
+        buttons = list(page_listing_more_buttons(self.en_blog_index, self.user))
+        translate_button = next(b for b in buttons if b.label == "Translate this page")
+        self.assertEqual(translate_button.priority, 35)
+
+    def test_page_header_buttons_priority(self):
+        buttons = list(page_header_buttons(self.en_blog_index, self.user))
+        translate_button = next(b for b in buttons if b.label == "Translate this page")
+        self.assertEqual(translate_button.priority, 55)
--- a/wagtail_localize/wagtail_hooks.py
+++ b/wagtail_localize/wagtail_hooks.py
@@ -181,10 +181,13 @@ def translation_buttons(page: Page, user, next_url=None, priority=60):
 
 
 def page_listing_more_buttons(page: Page, user, next_url=None):
+    # priority=35: after Delete (30), before Sort menu order (60)
     yield from translation_buttons(page, user, next_url, priority=35)
 
 
 def page_header_buttons(page: Page, user, view_name=None, next_url=None):
+    # view_name ("index" or "edit") is not used, both contexts share the same priority
+    # priority=55: after Delete (50), before Unpublish (60)
     yield from translation_buttons(page, user, next_url, priority=55)
 
 
--- a/wagtail_localize/wagtail_hooks.py
+++ b/wagtail_localize/wagtail_hooks.py
@@ -147,9 +147,8 @@ def register_submit_translation_permission():
     )
 
 
-def page_listing_more_buttons(page: Page, user, view_name=None, next_url=None):
+def translation_buttons(page: Page, user, next_url=None, priority=60):
     if not page.is_root() and user.has_perm("wagtail_localize.submit_translation"):
-        # If there's at least one locale that we haven't translated into yet, show "Translate this page" button
         has_locale_to_translate_to = Locale.objects.exclude(
             id__in=page.get_translations(inclusive=True)
             .exclude(alias_of__isnull=False)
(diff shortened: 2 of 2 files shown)


zanzo2003:
Thanks for tackling this, the root cause analysis in #911 is genuinely excellent, probably the best-documented issue in this repo. The fix is on the right track.

### What works well
Extracting shared logic into translation_buttons() is exactly the right pattern here. It makes the intent clear and eliminates the duplication that caused the original inconsistency. Registering separate wrappers per hook so each can carry its own priority is the correct architectural decision.

### A few things worth addressing before merge
- The view_name parameter gets silently dropped from page_header_buttons. The register_page_header_buttons hook passes view_name as either "index" or "edit" to distinguish between the Page Index and Page Editor contexts. In practice there's no regression right now since both get treated identically, but dropping the parameter makes the code harder to extend later (e.g. if we ever want different positioning in the editor vs the index). Worth either threading it through or leaving a comment explaining the decision.
- The chosen priority values (35 for listing, 55 for header) aren't documented anywhere. Given how much care went into mapping out the priority scales in #911, it would be really helpful to add a short inline comment explaining where each value lands relative to its neighbours, even just a one-liner like # places Translate after Delete (30), before Sort menu order (60). Without that context the numbers look arbitrary, and someone maintaining this a year from now will have to reverse-engineer it.
- The "Sync translated pages" button is also affected by this change (its priority now varies with the wrapper too), but that's not mentioned in the PR description. Worth a quick note just to make the changelog/review easier to follow.
- No tests is the main gap. The hook registration and button ordering are straightforward to cover — even a simple assertion that the buttons are registered with the expected priority per hook would prevent this class of regression from sneaking back in.

None of this is blocking in isolation, the core fix is sound. With a comment or two explaining the priority choices and a test, this is ready to go. Happy to help draft the test if that's useful.

rypptc:
Thanks for the thorough review. I've addressed all four points:

- Added a comment on `view_name` explaining it's received because Wagtail passes it, but not used since Page Index and Page Editor share the same priority. If they ever needed different values, `view_name` would be the way to distinguish them.
- Added inline comments documenting the priority values and their neighbors in each hook's scale.
- Updated the PR description to mention that "Sync translated pages" is also affected.
- Added two tests in `TestTranslationButtonPriority` verifying the expected priority per hook.

PR #930, opened 2026-06-03
Optimize PageIndex.from_database() to a constant query count

### LINKED ISSUE #927: Reduce query count in PageEntry.from_page_instance()
## Context

The `synctree.py` module synchronizes page trees across locales by creating aliases that point to the original content of each page.

To support this process, it builds a `PageIndex` — an in-memory index that maps each page to its translations and aliases across locales.

`PageIndex.from_database()` retrieves pages from the database and builds this index by creating one `Entry` object for each page.

`Entry.from_page_instance()` is responsible for extracting the data required for each entry, including basic page information, existing translations (`locales`), and existing aliases (`aliased_locales`).

## Current behavior

Building the index generates multiple database queries. The sequence is as follows.

`synchronize_tree()` calls `from_database()` to build the index:

```python
page_index = PageIndex.from_database().sort_by_tree_position()
```

Inside `from_database()`, a single query fetches all non-alias pages below the root:

```python
for page in Page.objects.filter(alias_of__isnull=True, depth__gt=1).only(
    *PageIndex.Entry.REQUIRED_PAGE_FIELDS
):
    pages.append(PageIndex.Entry.from_page_instance(page))
```

For each page in the loop, `from_page_instance()` issues up to 3 additional queries:

```python
@classmethod
def from_page_instance(cls, page):
    # Query 1 — fetches the parent page (only if depth > 2)
    parent_page = page.get_parent() if page.depth > 2 else None

    return cls(
        ...
        # Query 2 — fetches all locales where this page exists as a real translation
        list(
            Page.objects.filter(
                translation_key=page.translation_key,
                alias_of__isnull=True,
            ).values_list("locale", flat=True)
        ),
        # Query 3 — fetches all locales where this page exists as an alias
        list(
            Page.objects.filter(
                translation_key=page.translation_key,
                alias_of__isnull=False,
            ).values_list("locale", flat=True)
        ),
    )
```

## Problem

Given the structure described above, building the index for a site with N pages results in up to **3N + 1 database queries**.

## Possible approach

One possible direction would be to evaluate whether the data currently fetched inside the loop can instead be retrieved upfront using a small number of aggregated queries. This would reduce the number of queries from 3N + 1 to a constant number, regardless of how many pages the site has.

rypptc:
Closes #927

### Description

This PR includes several commits to document and fix the query behaviour of `PageIndex.from_database()`.

The first commit adds a characterization test showing that the number of queries grows with the number of pages in the tree.

The next commit optimizes `from_database()`: instead of issuing per-page queries inside the main loop, it does a single bulk query up front and builds lookup dictionaries (`locales`, `aliased_locales`, and parent resolution via `path` instead of `get_parent()`), plus `select_related("content_type", "locale")` to avoid per-page FK queries.

### Tests

The characterization test is updated to assert that the query count stays the same regardless of tree size (`assertEqual(small, large)`), instead of asserting that it grows. The 4 other existing tests in `test_synctree.py` still pass, confirming behaviour is preserved.

### AI usage

Used AI assistance to find the cause of the query growth and to draft the fix and its test. I reviewed, tested and understood every change.

### DIFF
--- a/wagtail_localize/synctree.py
+++ b/wagtail_localize/synctree.py
@@ -63,31 +63,26 @@ class PageIndex:
         ]
 
         @classmethod
-        def from_page_instance(cls, page):
+        def from_page_instance(
+            cls, page, locales_by_key, aliased_locales_by_key, translation_key_by_path
+        ):
             """
             Initialises an Entry from the given page instance.
             """
             # Get parent, but only if the parent is not the root page. We consider the
             # homepage of each langauge tree to be the roots
-            parent_page = page.get_parent() if page.depth > 2 else None
+            parent_path = page.path[: -Page.steplen] if page.depth > 2 else None
+            parent_translation_key = (
+                translation_key_by_path.get(parent_path) if parent_path else None
+            )
 
             return cls(
                 page.content_type,
                 page.translation_key,
                 page.locale,
-                parent_page.translation_key if parent_page else None,
-                list(
-                    Page.objects.filter(
-                        translation_key=page.translation_key,
-                        alias_of__isnull=True,
-                    ).values_list("locale", flat=True)
-                ),
-                list(
-                    Page.objects.filter(
-                        translation_key=page.translation_key,
-                        alias_of__isnull=False,
-                    ).values_list("locale", flat=True)
-                ),
+                parent_translation_key,
+                locales_by_key.get(page.translation_key, []),
+                aliased_locales_by_key.get(page.translation_key, []),
             )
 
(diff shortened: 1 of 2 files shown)


rypptc:
Closing in favour of #953, which replaces this.

This branch predates the src/ layout from #941, so its paths no longer apply, and it was written before the benchmark harness existed. #953 carries the same change to synctree.py, a smaller regression test that asserts the query count does not grow rather than pinning an absolute number, and before/after measurements from the harness proposed in #946.

PR #921, opened 2026-05-13
Wagtail 7.4 LTS maintenance

nickmoreton:
## Summary

- Align the supported version matrix with the current Wagtail / Django / Python support windows: drop Wagtail 6.3 (EOL 2026-05-01), keep 7.0 LTS / 7.2 / 7.3, add 7.4 LTS. Wagtail 7.1 is excluded — it was a non-LTS feature release that has already aged out of support. Minimum Wagtail is now 7.0.
- Drop Django 4.2 (EOL April 2026) — minimum Django is now 5.1. This also dodges the wagtail-modeladmin 2.3.0 regression that only fires on Django <5.0.
- Refresh the tox matrix accordingly (Django 5.1 / 5.2 / 6.0 across Python 3.10–3.14) and update classifiers, README, and CHANGELOG.
- Remove stale `WAGTAIL_VERSION` / `DJANGO_VERSION` compatibility guards that no longer cover any supported version.
- Fix a Wagtail 7.3+ regression in `Translation.save_target()` that raised `IntegrityError` on `(translation_key, locale_id)` when translating snippets with nested `TranslatableMixin` cluster grandchildren. Wagtail's `CopyForTranslationAction` only re-locales the top tier of the returned cluster; the workaround walks the cluster after `copy_for_translation` and re-assigns the target locale on every nested descendant.
- Modernise non-production tooling: Python docs/testing extras, pre-commit hook revs (ruff 0.15.12, `ruff-check` id), GitHub Actions majors, and the Node build stack (webpack 5, TypeScript 5, drop `file-loader` in favour of webpack 5 asset modules).
- Fix the tox-gh-actions postgres factor mapping (`postgres → postgres15`) so the matrix entries actually run instead of falling back to the default `py` env.

Commits, in order:

1. `dc3b7cc` Add Wagtail 7.4 LTS support, drop Wagtail 6.3
2. `f32bcd0` Remove Wagtail 7.1 from the supported version range
3. `725ea4a` Update Python documentation and testing dev dependencies
4. `c24eb4f` Bump pre-commit hook versions
5. `f17ed81` Bump GitHub Actions to current majors
6. `b5d5541` Bump Node build tooling to webpack 5 / TypeScript 5
7. `410ff23` Fix tox-gh-actions postgres env factor mapping
8. `5947e94` Fix nested-cluster locale on snippet translation save
9. `ac77fdf` Drop Django 4.2 support

## Deliberately deferred

[...]

PR #934, opened 2026-06-17
Fix #922: propagate target locale to all descendants when translating nested snippets

### LINKED ISSUE #922: Tests fail to run under some conditions (?)
### Issue summary

There’s a single failure in the tests when I run them locally:

```
======================================================================
ERROR: test_edit_nested_snippet_translation (wagtail_localize.tests.test_edit_translation.TestGetEditTranslationView.test_edit_nested_snippet_translation)
----------------------------------------------------------------------
  File "/nix/store/llkagd3zdsc5yzyvnrqf6iisnfwadwdc-python3.12-django-5.2.13/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/llkagd3zdsc5yzyvnrqf6iisnfwadwdc-python3.12-django-5.2.13/lib/python3.12/site-packages/django/db/backends/sqlite3/base.py", line 360, in execute
    return super().execute(query, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlite3.IntegrityError: UNIQUE constraint failed: wagtail_localize_test_subnavigationlink.translation_key, wagtail_localize_test_subnavigationlink.locale_id
```

The problem occurs with Wagtail 7.4, 7.3 and 7.2. I couldn’t figure out what causes it to fail.

Also Wagtail 7.3 is not covered by the Tox matrix.

### Steps to reproduce

1. Run `python testmanage.py test` locally

### Additional information

_No response_

### Can be reproduced

Not confirmed

### Technical details

- Python version: 3.12
- Django version: 5.2.13
- Wagtail version: 7.4

Python environment:

```
anyascii==0.3.3
asgiref==3.11.0
beautifulsoup4==4.14.3
certifi==2026.1.4
charset-normalizer==3.4.4
click==8.3.1
croniter==6.0.0
defusedxml==0.8.0rc2
dj-database-url==0.0.0
Django==5.2.13
django-filter==25.2
django-modelcluster==6.5
django-permissionedforms==0.1
django-rq==4.1.0
django-stubs-ext==5.2.9
django-taggit==6.1.0
django-tasks==0.12.0
django-treebeard==4.8.0
djangorestframework==3.16.1
draftjs_exporter==5.1.0
et_xmlfile==2.0.0
filetype==1.2.0
freezegun==1.5.5
google==3.0.0
idna==3.11
laces==0.1.2
modelsearch==1.3.1
openpyxl==3.1.5
pillow==12.2.0
pillow_heif==1.3.0
polib==1.2.0
python-dateutil==2.9.0.post0
pytz==2026.1.post1
redis==7.4.0
requests==2.33.1
rq==2.8.0
setuptools==80.10.1
six==1.17.0
soupsieve==2.8.3
sqlparse==0.5.5
telepath==0.3.1
typing_extensions==4.15.0
urllib3==2.6.3
wagtail==7.4
wagtail-modeladmin==2.3.0
Willow==1.12.0
```

### Working on this

_No response_

rypptc:
Fixes #922

### Description

Translating nested snippets triggers a recurring problem when there are more than two levels of nesting, for example:

Header > NavigationLink > SubNavigationLink

The issue happens inside `copy_synchronised_fields`, specifically on this line:

```python
child_object_map = source.copy_child_relation(field.name, target)
```

`copy_child_relation` (in `modelcluster.models`) internally calls `copy_cluster()` on each direct child that is a `ClusterableModel`. `copy_cluster()` deep-copies the whole tree, including the grandchildren, and returns its own map of what it copied. But `copy_child_relation` never merges that nested map into the one it's building — it only records the direct child itself (the `NavigationLink`), then discards the nested map (`child_object, _ = child_object.copy_cluster()`). So `child_object_map` ends up with just the first level: the tree is copied in full, but the grandchildren (`SubNavigationLink`) never make it into the map.

Then, when the locales are updated, the loop only iterates over `child_object_map`, so only the first level is updated:

```python
child_object.locale = target.locale
```

The `NavigationLink` is moved to the target locale, but the `SubNavigationLink` stays in the source one. When the tree is saved, that record collides with the existing original and violates the `(translation_key, locale_id)` constraint, resulting in:

```
IntegrityError: UNIQUE constraint failed:
  wagtail_localize_test_subnavigationlink.translation_key, wagtail_localize_test_subnavigationlink.locale_id
```

### Solution

Since that map doesn't reach the grandchildren, instead of relying on it the proposal is to walk the tree after the copy and reassign the locale to every descendant.

For this, a `_update_descendants_locale` helper is added, which recursively descends through the child relations and reassigns the target locale to every descendant that is a `TranslatableMixin`. It is called from `copy_synchronised_fields`, right after updating the direct child's locale, when that child is a `ClusterableModel`.

Because it lives in `copy_synchronised_fields`, the fix covers the three paths that go through that function: the first translation, the re-sync (when the translation already existed) and the preview (`get_ephemeral_translated_instance`).

### Verification

<details>
<summary>Step-by-step verification (bug + fix, click to expand)</summary>

This is not a copy of the actual package code — it's a simplified reproduction of the same mechanism, built with the real underlying APIs, to make the root cause and the fix visible step by step. Specifically: `copy_cluster()` and the `child_object_map` it returns are the real modelcluster mechanism, called directly. The locale-update step iterates that map exactly as the loop over `child_object_map.items()` in `copy_synchronised_fields` does today, omitting only the primary-key recycling and the unsaved-children branch, which aren't needed to show the bug. `_update_descendants_locale` is the exact function proposed in this PR (see Solution above).

Run in `python testmanage.py shell` (from the wagtail-localize repo root):

```python
from wagtail.models import Locale
from wagtail_localize.test.models import Header, NavigationLink, SubNavigationLink

fr, _ = Locale.objects.get_or_create(language_code="fr")

h = Header.objects.create(name="H", navigation_links=[
    NavigationLink(label="Nav", sub_navigation_links=[SubNavigationLink(label="SubNav")])
])
```

Before copying the whole `Header`, we reproduce by hand the exact line that `modelcluster` runs internally when it copies a `NavigationLink` (`copy_child_relation`, `modelcluster/models.py:352`):

```python
nav = h.navigation_links.first()
nav_copy, _ = nav.copy_cluster()
_
```

Output:

```
{(<ManyToOneRel: wagtail_localize_test.subnavigationlink>, 17): <SubNavigationLink: SubNavigationLink object (None)>}
```

`_` is the map `copy_cluster()` produces when copying the `NavigationLink` — it does include the `SubNavigationLink`. In the real code, nothing reads `_` after this line; it's lost right here.

Now we copy the whole `Header` — `copy_cluster()` reaches the same `copy_child_relation` that `copy_synchronised_fields` calls directly, so `header_map` below has the same shape:

```python
copy, header_map = h.copy_cluster()
header_map
```

Output:

```
{(<ManyToOneRel: wagtail_localize_test.navigationlink>, 17): <NavigationLink: Nav>}
```

This is the map that does survive — the same shape returned by `source.copy_child_relation(field.name, target)` in `copy_synchronised_fields`. It only has the `NavigationLink`; the `SubNavigationLink` map we saw above never gets merged into it.

```python
copy.pk = None

copy.locale.language_code
copy.navigation_links.first().locale.language_code
copy.navigation_links.first().sub_navigation_links.first().locale.language_code
```

Output — raw copy: whole tree copied, locale untouched:

```
'en'
'en'
'en'
```

Applying the locale update that `copy_synchronised_fields` does today: iterating `header_map` (the same map `source.copy_child_relation(field.name, target)` returns). The root's locale is already set before this function runs, by `copy_for_translation()` or the existing target instance. Because `header_map` only holds the direct child, this loop can't reach the `SubNavigationLink`:

```python
copy.locale = fr
for (_child_relation, _source_pk), child in header_map.items():
    child.locale = fr

copy.locale.language_code
copy.navigation_links.first().locale.language_code
copy.navigation_links.first().sub_navigation_links.first().locale.language_code
```

Output — `NavigationLink` fixed, `SubNavigationLink` left behind:

```
'fr'
'fr'
'en'
```

The grandchild isn't lost — it's reachable through the tree and gets saved, which is exactly why it collides:

```python
orig_sub = h.navigation_links.first().sub_navigation_links.first()
copy_sub = copy.navigation_links.first().sub_navigation_links.first()

orig_sub.translation_key == copy_sub.translation_key
orig_sub.locale.language_code, copy_sub.locale.language_code

copy.save()
```

Output — same `translation_key`, same `locale` as the original English row: the forbidden duplicate pair, and the resulting error:

```
True
('en', 'en')

django.db.utils.IntegrityError: UNIQUE constraint failed:
  wagtail_localize_test_subnavigationlink.translation_key, wagtail_localize_test_subnavigationlink.locale_id
```

Now with `_update_descendants_locale` (this PR's fix), walking the tree instead of the one-level map. Using a fresh source tree here — the failed `save()` above is not atomic, so it already wrote the `Header` row before crashing on the grandchild, and reusing `h` would collide on that leftover row instead of showing the fix:

```python
from modelcluster.models import ClusterableModel, get_all_child_relations
from wagtail.models import TranslatableMixin

def _update_descendants_locale(obj, target_locale):
    for child_relation in get_all_child_relations(obj):
        for descendant in getattr(obj, child_relation.get_accessor_name()).all():
            if isinstance(descendant, TranslatableMixin):
                descendant.locale = target_locale
            if isinstance(descendant, ClusterableModel):
                _update_descendants_locale(descendant, target_locale)

h2 = Header.objects.create(name="H2", navigation_links=[
    NavigationLink(label="Nav", sub_navigation_links=[SubNavigationLink(label="SubNav")])
])

copy2, header_map2 = h2.copy_cluster()
copy2.pk = None
copy2.locale = fr
for (_child_relation, _source_pk), child in header_map2.items():
    child.locale = fr
    _update_descendants_locale(child, fr)

copy2.navigation_links.first().sub_navigation_links.first().locale.language_code

copy2.save()
```

Output — the grandchild is reached this time, and the save succeeds:

```
'fr'
# save: no error
```

After the map loop sets each direct child's locale, `_update_descendants_locale` recurses into anything that's a `ClusterableModel`, so every level gets the target locale, not just the first one.

</details>

### Tests

A regression test is added for the re-sync case: translating a snippet whose nested tree already exists in the target locale. `save_target()` is called twice (the first creates the FR tree, the second re-copies it) and it checks that a single `SubNavigationLink` still exists in FR, without raising the `IntegrityError`.

This is a case the existing test (`test_edit_nested_snippet_translation`) didn't cover, since it only verified the first translation.

### Note on #921

#921 includes a fix for this bug, but there are some cases it doesn't cover, such as re-sync and preview. There may be overlap between both PRs.

### AI usage

AI-assisted: an LLM was used to help explore the codebase, draft the fix and the regression test, and write this description. All changes were reviewed, tested, and understood.

### DIFF
--- a/wagtail_localize/fields.py
+++ b/wagtail_localize/fields.py
@@ -233,6 +233,19 @@ def get_translatable_fields(model):
         return translatable_fields
 
 
+def _update_descendants_locale(obj, target_locale):
+    """
+    Recursively sets the locale on every TranslatableMixin descendant of obj.
+    """
+    for child_relation in get_all_child_relations(obj):
+        accessor = child_relation.get_accessor_name()
+        for descendant in getattr(obj, accessor).all():
+            if isinstance(descendant, TranslatableMixin):
+                descendant.locale = target_locale
+            if isinstance(descendant, ClusterableModel):
+                _update_descendants_locale(descendant, target_locale)
+
+
 def copy_synchronised_fields(source, target):
     """
     Copies data in synchronised fields from the source instance to the target instance.
@@ -274,6 +287,10 @@ def copy_synchronised_fields(source, target):
                                     child_object.translation_key
                                 )
                                 child_object.locale = target.locale
+                                if isinstance(child_object, ClusterableModel):
+                                    _update_descendants_locale(
+                                        child_object, target.locale
+                                    )
                         else:
                             child_object = child_objects
 
@@ -281,6 +298,8 @@ def copy_synchronised_fields(source, target):
                                 child_object.translation_key
                             )
                             child_object.locale = target.locale
+                            if isinstance(child_object, ClusterableModel):
+                                _update_descendants_locale(child_object, target.locale)
 
                 else:
                     source.copy_child_relation(field.name, target)
--- a/wagtail_localize/fields.py
+++ b/wagtail_localize/fields.py
@@ -237,6 +237,10 @@ def _update_descendants_locale(obj, target_locale):
     """
     Recursively sets the locale on every TranslatableMixin descendant of obj.
     """
+    # FIXME: this is a workaround for an upstream issue in Wagtail:
+    # https://github.com/wagtail/wagtail/issues/14425 This might be removed in
+    # the future if the upstream issue is fixed, but for now we need to ship our
+    # own workaround.
     for child_relation in get_all_child_relations(obj):
         accessor = child_relation.get_accessor_name()
         for descendant in getattr(obj, accessor).all():
--- a/wagtail_localize/models.py
+++ b/wagtail_localize/models.py
@@ -35,7 +35,6 @@ from django.utils.translation import gettext_lazy
 from modelcluster.fields import ParentalKey
 from modelcluster.models import (
     ClusterableModel,
-    get_all_child_relations,
     get_serializable_data_for_fields,
     model_from_serializable_data,
 )
@@ -269,28 +268,6 @@ class MissingRelatedObjectError(Exception):
         self.locale = locale
 
         super().__init__()
-
-
-def _set_locale_on_nested_translatable_children(instance, locale):
-    """
-    Recursively walk every ParentalKey child relation on a cluster and set
-    `locale` on each `TranslatableMixin` descendant.
-
-    Wagtail's `copy_for_translation` only re-locales the direct entries in the
-    top-level `child_object_map`, leaving grandchildren of nested
-    `ClusterableModel` children at the source locale. Without this fixup, the
-    `(translation_key, locale_id)` unique constraint on those grandchildren
-    collides with the source rows on save.
-    """
-    if not isinstance(instance, ClusterableModel):
-        return
-
-    for child_relation in get_all_child_relations(instance):
-        accessor_name = child_relation.get_accessor_name()
-        for child in getattr(instance, accessor_name).all():
-            if isinstance(child, TranslatableMixin):
-                child.locale = locale
-            _set_locale_on_nested_translatable_children(child, locale)
 
 
 class TranslationSourceQuerySet(models.QuerySet):
@@ -772,14 +749,6 @@ class TranslationSource(models.Model):
                 )
             else:
                 translation = original.copy_for_translation(locale)
-
-            # Wagtail's `copy_for_translation` only switches the locale of
-            # direct child objects in the returned cluster. Grandchildren of
-            # nested `ClusterableModel` children keep the source locale, which
-            # collides on the `(translation_key, locale_id)` unique constraint
-            # when saved. Walk the cluster and re-assign the target locale on
-            # every nested translatable child.
-            _set_locale_on_nested_translatable_children(translation, locale)
 
             created = True
 
--- a/wagtail_localize/tests/test_edit_translation.py
+++ b/wagtail_localize/tests/test_edit_translation.py
@@ -1743,6 +1743,7 @@ class TestGetEditTranslationView(EditTranslationTestData, TestCase):
         self.assertEqual(response.status_code, 200)
 
     def test_resync_nested_snippet_translation_keeps_locale(self):
+        # This tests the implementation of `wagtail_localize.fields._update_descendants_locale`
         # Re-syncing a translation whose nested tree already exists must keep the
         # deeply-nested SubNavigationLink in the target locale (it is copied
         # again but its locale must still be updated).
--- a/wagtail_localize/tests/test_edit_translation.py
+++ b/wagtail_localize/tests/test_edit_translation.py
@@ -1742,6 +1742,32 @@ class TestGetEditTranslationView(EditTranslationTestData, TestCase):
         )
         self.assertEqual(response.status_code, 200)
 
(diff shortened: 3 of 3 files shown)


Stormheg (approved):
Thank you @rypptc – this seems like a good fix to me. Ideally, this should be fixed in upstream Wagtail itself, but for the time being I'm fine shipping this with wagtail-localize.

What do you think would happen when this is fixed in Wagtail? Would this workaround code conflict with whatever Wagtail's fix will do? (not something I'm super worried about, we can patch wagtail-localize when it comes to that 👍 )

Maybe it would be wise to add some a `FIXME` code comment with a reference to your upstream wagtail issue. That way it is easier for future maintainers to identify if the workaround is still necessary.

rypptc:
Thank you very much for the review! I don't think a fix in Wagtail would reach this. `copy_for_translation` only runs on the first translation, and even there `copy_synchronised_fields` replaces the child objects it produced with new copies from the source (`fields.py:279-280`). So, the fix in Wagtail won't affect the result in wagtail-localize.

PR #940, opened 2026-08-13
Revised: Wagtail 7.4 support

Stormheg:
Supersedes #921 

- Undoes the node toolchain changes, as those are better handled in a separate PR
- Updates the tox matrix to current Wagtail and Django versions
- Fix mkdocs build, this needs a major overhaul as things are quite dated. Not just the tooling, but screenshots as well.


mkdocs.yml
```diff
@@ -36,11 +36,8 @@ plugins:
           options:
             show_root_toc_entry: false
             show_signature_annotations: true
-          setup_commands:
-            - import os
-            - import django
-            - os.environ.setdefault("DJANGO_SETTINGS_MODULE", "wagtail_localize.test.settings")
-            - django.setup()
+          paths:
+            - wagtail_localize
```
Stormheg, line 40: `setup_commands` doesn't exist anymore in newer versions of mkdocstrings. This is an AI-assisted alternative that seems to have the same effect.

Tested by checking if the docstrings still render, which they seem to do alright.

pyproject.toml
```diff
@@ -43,17 +41,18 @@ google = [
...
-    "mkdocstrings[python]==0.22.0",
-    "mkdocs-autorefs>=0.4.0,<0.5",
-    "mkdocs-include-markdown-plugin>=4.0.4,<5",
+    "mkdocs==1.5.3",
+    "mkdocs-material>=9.2,<9.3",
+    "mkdocstrings==0.28.3",
+    "mkdocstrings-python==1.16.12",
+    "mkdocs-autorefs>=1.0,<2",
+    "mkdocs-include-markdown-plugin>=7.0,<8",
     "pygments>=2.15,<2.16",
```
Stormheg, line 50: The docs seem to build alright with these constraints. Severe software rot here, need to revisit this. Maybe switch over to different documentation generator tooling?

PR #929, opened 2026-06-01
Add demo site and adopt src/ layout

rypptc:
Fixes #928

### Description

Adds a `demo/` folder based on [cookiecutter-wagtail-package/demo](https://github.com/wagtail/cookiecutter-wagtail-package/tree/main/demo), adapted to showcase wagtail-localize features for local development and as a reference for potential users of the package.

Changes from the original cookiecutter demo:
- Added `wagtail_localize` and `wagtail_localize.locales` to `INSTALLED_APPS`
- Enabled `WAGTAIL_I18N_ENABLED = True` with EN/FR content languages
- Configured `LocaleMiddleware` and `i18n_patterns` in URLs
- Added `demo/` to `.eslintignore`

Adopted a `src/`-based layout to separate the test suite and the demo from the installable package. The `wagtail_localize` package now lives under `src/wagtail_localize/`, keeping only production code — the public import is unchanged (`import wagtail_localize`).

The test suite, previously scattered across `wagtail_localize/tests/`, `segments/tests/`, `machine_translators/tests/`, etc., is now unified under a single top-level `tests/` package. The test app used by the suite moved from `wagtail_localize/test/` to `tests/testapp/`, keeping the existing app label (`wagtail_localize_test`) so table names and migrations are unaffected.

The `src/` layout and the top-level `tests/`/`demo/` split follow the structure used by `cookiecutter-wagtail-package`. The test app specifically was moved to `tests/testapp/`, which differs from the cookiecutter's own convention of keeping the test app inside the package.

The test app and the demo were kept separate rather than unified. They serve different purposes: the test app's models are designed to exercise edge cases (nested snippets, custom fields, orderable relations, etc.), while the demo aims to showcase a realistic, navigable site. `cookiecutter-wagtail-package` itself keeps these two separate as well.

All configuration referencing the old paths was updated accordingly (CI, coverage, packaging, build tooling, docs).

Validated that the restructure doesn't change behaviour: the full test suite (444 tests) still passes with default, label-less discovery — the same way tox/CI invoke it — with only the pre-existing, unrelated #922 failure. `flit build` still produces a clean wheel/sdist (no test code or static source included).

To run the demo:

```bash
cd demo
python manage.py migrate
python manage.py load_initial_data
python manage.py runserver
```

Admin credentials: `admin` / `changeme`

### AI usage

[...]

Issue #928, opened 2026-06-01
Add demo site for wagtail-localize


rypptc:
A new `demo` folder as a copy of https://github.com/wagtail/cookiecutter-wagtail-package/tree/main/demo, to be used as a reference implementation of wagtail-localize. For local development and for potential users of the package.

thibaudcolas:
Extra improvements after the first version:

- More advanced localize patterns
- More locales- 
- DeepL machine translator
- More demo unit tests within the same CI job as current PR
- Language picker: wait on https://github.com/wagtail/bakerydemo/pull/760

PR #941, opened 2026-08-16
Adopt a src/ layout and move the test suite to tests/

rypptc:
This is the `src/` layout half of #929, which combined two changes. The demo site is in #942.

#929 started out as the PR for the demo site (#928). @zerolab suggested there that the project switch to an `src/`-based structure, and the PR ended up carrying both changes at once.

### Description

The `src/` layout separates what ships in the installable package from what doesn't. It is the structure [cookiecutter-wagtail-package](https://github.com/wagtail/cookiecutter-wagtail-package) uses.

The package moves to `src/wagtail_localize/`; `import wagtail_localize` is unchanged. The test suite is unified under a top-level `tests/`, and the test app moves to `tests/testapp/` keeping its app label, so no new migrations are needed.

Most of the diff is files moving. The only content changes are updated imports in the tests, and the config files that pointed at the old paths.

Two decisions came out of rebasing onto current `main` and are worth a look. In `mkdocs.yml`, the recent docs fix replaced `setup_commands` with `paths`; I kept that block and pointed `paths` at `src`, and the docs still build. In the test settings, `STATICFILES_DIRS` pointed at a directory that no longer exists after the move, and was redundant with `AppDirectoriesFinder` anyway, so I removed it.

Verified locally: the test suite, the pre-commit hooks and the docs build all pass.

### AI usage

Used AI assistance to reapply the original commits onto current `main` and to review the result, and to draft this description. I ran the tests, hooks and docs build locally.


Stormheg (approved):
Yep, all good to me.


I wanted to make a note of how the `tests` directory should be included with [`sdist`](https://pydevtools.com/handbook/reference/sdist/) builds since its a general Python ecosystem expectation that sdists contain everything needed to build from source and run the tests. However, I see the that is currently explicitly not the case (static_src should also be included but its not). Will make a follow up issue about that.

https://github.com/wagtail/wagtail-localize/blob/28abec1778a84e43110e2a2148bfcf1225850144/pyproject.toml#L71-L94

PR #942, opened 2026-08-16
Add a demo site for testing and development

### LINKED ISSUE #928: Add demo site for wagtail-localize

rypptc:
Closes #928. This is the demo site half of #929, which combined two changes. The `src/` layout is in #941.

### Description

The demo is a reference implementation of wagtail-localize: a site for local development, and for anyone evaluating the package. The `demo/` folder is based on the demo in [cookiecutter-wagtail-package](https://github.com/wagtail/cookiecutter-wagtail-package/tree/main/demo), adapted for wagtail-localize.

The demo and the test app are kept separate. The test app's models exist to exercise edge cases; the demo aims to be a realistic, navigable site. `cookiecutter-wagtail-package` keeps them separate too.

The review of #929 suggested that `tests/testapp/` could double as the demo app, so that both would run off the same models, ensuring full coverage. I kept them separate for the reason above, but tried to address the coverage concern another way: this PR adds a CI job that runs the demo's own checks and tests.

Running the tests and starting the demo turned up two bugs. `LANGUAGE_CODE` was left at Django's `en-us` default while the configured languages are English and French, so page URLs carried a prefix that doesn't resolve and the demo's tests returned 404. And `load_initial_data` could not run at all: the demo's `.gitignore` was unintentionally excluding the images the command needs, so they were never committed.

To run the demo:

```bash
cd demo
python manage.py migrate
python manage.py load_initial_data
python manage.py runserver
```

User `admin`, password `changeme`.

The demo doesn't yet demonstrate much of wagtail-localize: no language switcher, no translated content in the initial data, no translatable snippets. Those are planned as a follow-up.

### AI usage

Used AI assistance to separate the demo commits from the layout ones and to review the result, to diagnose the two bugs, and to draft this description. I ran every command in this description locally.


demo/demo/settings/base.py
```diff
@@ -0,0 +1,206 @@
...
+
+USE_I18N = True
+
+WAGTAIL_I18N_ENABLED = True
+
+USE_TZ = True
+
+WAGTAIL_CONTENT_LANGUAGES = LANGUAGES = [
+    ("en", "English"),
+    ("fr", "French"),
```
thibaudcolas, line 137: we want more!

thibaudcolas (approved):
Major improvement! See #928 for follow-ups. Also need to investigate that userbar bug

Issue #938, opened 2026-07-09
Benchmark test suite / tooling


thibaudcolas:
### Problem statement

Follow-up to @rypptc’s [wl-benchmark](https://github.com/rypptc/wl-benchmark) - we want to bring this kind of benchmark harness into the project for long-term use.

### Proposed solution

Requirements:

- Easy to integrate with existing code
- Can be run on demand locally or in CI
- Can track query count / query execution time
- Some reporting? console or HTML?

This can probably be achieved more easily with a combination of tools rather than necessarily "one to rule them all"

Packages/tools suggested by @allcaps:

- https://github.com/jazzband/django-silk -- very useful to know about but not sure it’s the best fit here
- https://adinhodovic.github.io/django-o11y/ -- looks a bit too complex?
- https://pytest-django.readthedocs.io/en/latest/helpers.html#django-assert-num-queries -- this or Django’s built-in `assertNumQueries ` definitely a must
- https://github.com/xelixdev/django-eagle -- looks very neat even if limited
- https://pypi.org/project/pytest-subtests/

### Working on this

Assigning to @rypptc

allcaps:
These are profiling, telemetry, and test utilities. Each have their own use-case. I see the profiling and telemetry as development tooling. We night use them to find weak spots. They don't need to be permanently added to the package.

Maybe we should go for the smallest change with highest value? Only add Django unit test AssertNumQueries to the test suite? Create regression tests, that guard against performance degradation.

rypptc:
I've created a couple of branches to start trying out the different tools. I based them on PR #929 (demo site + `src/` layout), so that anything added under `benchmarks/` stays outside the package and doesn't affect it.

The first branch sets up the shared base for experimenting with the different tools:
https://github.com/rypptc/wagtail-localize/tree/eval/scaling-tooling-base

What's in it, under `benchmarks/`:

- `env.py` — boots Django (settings + database) so the benchmark runs as a plain script, without going through the test runner.
- `seed.py` — builds the sample content the measurements need: a small tree of pages, some snippets, and the locales.
- `flows.py` — the catalog of translation operations we measure (creating translations, syncing the locale tree, refreshing segments, saving a target, the submit/update admin views, and the report and snippet API). It only describes *what* to measure and imports no measurement tool, so any tool can reuse the same flows.

The second branch experiments with django-query-doctor:
https://github.com/rypptc/wagtail-localize/tree/eval/scaling-tooling-django-query-doctor

[...]

PR #946, opened 2026-08-17
Add performance benchmark harness

rypptc:
## Summary

Adds a performance benchmark harness for reproducible measurement and diagnosis of translation flows, and for validating optimizations to them later. Refs wagtail/wagtail-localize#938.

## What it does

The harness runs real translation operations in an isolated process against a clean database, and records query count and time.

The harness defines a catalogue of eight named flows, each representing a real translation operation, such as submitting a page for translation, translating a snippet, syncing a page tree, or refreshing segments. A flow can declare two input sizes, so results show how its cost grows rather than only its total, and it can be repeated to summarise timing variance. `--list` prints the catalogue, including what each flow covers.

```console
$ python benchmarks/run.py core_page_index --repeat 5
core_page_index [small]  119 queries  25.5 ms median (min 23.8, max 29.2, 5 repeats)  24 indexed_pages (expected 24)
core_page_index [large]  319 queries  68.1 ms median (min 64.6, max 80.1, 5 repeats)  64 indexed_pages (expected 64)
```

Query counts are identical across repeats; time varies, so it is reported as median, minimum, and maximum. Fixture construction and verification stay outside the measured section. The run fails if its observed workload differs from the catalogue's expected workload, because measuring a scenario that did not build would describe something else. `--json PATH` additionally writes the run with its commit, branch, and dependency versions, so any number can be reproduced. The optimisation for this flow is deliberately out of scope for this PR and is tracked separately.

`benchmarks/run_query_doctor.py` is a separate, manual command that runs the same flow under Django Query Doctor to explain *why* a number is what it is, including a fix suggestion when Query Doctor has one. `run.py` does not import it, does not know it exists, and runs unchanged in an environment where the tool is not installed.

## Developer tooling, not a CI gate

The harness is run by hand when investigating or validating a change. Query-count regression protection stays where it already lives: `assertNumQueries` / `CaptureQueriesContext` in targeted product tests, added per fix.

Concretely, in CI:

- the harness's tests run in the existing matrix, with no new dependencies: structural tests for the runner and catalogue, plus one in-process smoke test that builds the fixture and runs the cheapest flow, so a change to the test app's models that breaks the fixture fails in CI rather than on the next manual run;
- the benchmarks themselves do not run;
- Query Doctor is not involved at all — its test suite passes without it installed.

## Scope

- Adds developer tooling under `benchmarks/`; kept in the repository but shipped in neither the wheel nor the sdist. Since #945 the sdist is built from an explicit `include` list in `[tool.flit.sdist]`, and `benchmarks/` is not on it.
- Adds tests for the harness and for the Query Doctor command.
- Does not modify `src/wagtail_localize`, runtime dependencies, or product behaviour.
- Benchmark and Query Doctor commands run only when explicitly invoked.

### AI usage

AI-assisted: an LLM helped design and implement the harness (catalogue, runner, and tests) iteratively across several sessions, and helped draft this description. All changes were reviewed, tested, and understood.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,11 @@ or, you can run them for a specific environment `tox -e python3.13-django5.2-wag
 
 To run the test app interactively, use `tox -e interactive`, visit `http://127.0.0.1:8020/admin/` and log in with `admin`/`changeme`.
 
+### Performance benchmarks
+
+See [`benchmarks/README.md`](benchmarks/README.md) for the benchmark catalog,
+methodology, and commands for reproducible before/after measurements.
+
 ## Support
 
 For support, please use [GitHub Discussions](https://github.com/wagtail/wagtail-localize/discussions) or ask a question on the `#multi-language` channel on [Wagtail's Slack instance](https://wagtail.org/slack/).
--- a/README.md
+++ b/README.md
@@ -108,8 +108,9 @@ To run the test app interactively, use `tox -e interactive`, visit `http://127.0
 
 ### Performance benchmarks
 
-See [`benchmarks/README.md`](benchmarks/README.md) for the benchmark catalog,
-methodology, and commands for reproducible before/after measurements.
+See [`benchmarks/README.md`](https://github.com/wagtail/wagtail-localize/blob/main/benchmarks/README.md)
+for the benchmark catalog, methodology, and commands for reproducible
+before/after measurements.
 
 ## Support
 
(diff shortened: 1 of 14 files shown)


thibaudcolas (approved):
- Consider how / whether we would want to use this with PostgreSQL?
- Be great to run a subset of this in CI
- Move django-query-doctor to be managed with other testing dependencies
- Nice to have tests for benchmark suite!

PR #947, opened 2026-08-18
Bump Node build tooling to webpack 5 / TypeScript 5

thibaudcolas:
Cherry-pick of [torchbox-forks/wagtail-localize@b5d5541](https://github.com/torchbox-forks/wagtail-localize/commit/b5d554193182a6a1b3ec1178adaed9dc3ffbbe12), initially in #921

> […]
> React / react-dom / styled-components intentionally left at 16.x / 5.x — bumping React types to 18 surfaces a known incompatibility with @types/styled-components 5, deferred to a dedicated change.

---

Also adds a frontend job to CI that installs Node dependencies and runs `npm run build`, so this tooling is exercised on every push/PR (previously only done at publish time).


.node-version
```diff
@@ -1 +1 @@
-14
+22
```
Stormheg: Can we go all the way to v26?

thibaudcolas: certainly!

Stormheg:
@thibaudcolas I'm a bit nervous about the tooling changes. Can you confirm the translation editor still works? Other than that, I'm okay with this.



Wreck-it-Ralp says: I'm gonna ship it!

thibaudcolas:
@Stormheg as far as I could tell it’s working fine yes!

allcaps (approved):

PR #948, opened 2026-08-18
Fix warning about inconsistent ordering

Stormheg:
### Description

Fixes this warning in the testsuite output

	UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'wagtail_localize.models.Translation'> QuerySet.

### AI usage

None


zerolab (approved):
works for me =)

PR #955, opened 2026-08-21
Handle removed fields when editing translations with draft source

### LINKED ISSUE #590: `FieldDoesNotExist` error when editing translation with draft source after migration removes field
## Traceback

```
Internal Server Error: /admin/pages/10/edit/
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/django/db/models/options.py", line 672, in get_field
    return self.fields_map[field_name]
KeyError: 'foo_description'

During handling of the above exception, another exception occurred:

  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/django/views/decorators/cache.py", line 62, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail/admin/urls/__init__.py", line 161, in wrapper
    return view_func(request, *args, **kwargs)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail/admin/auth.py", line 205, in decorated_view
    return view_func(request, *args, **kwargs)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/django/views/generic/base.py", line 84, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail/admin/views/pages/edit.py", line 341, in dispatch
    response = self.run_hook("before_edit_page", self.request, self.page)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail/admin/views/generic/hooks.py", line 16, in run_hook
    result = fn(*args, **kwargs)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail_localize/wagtail_hooks.py", line 259, in before_edit_page
    return edit_translation.edit_translation(request, translation, page)
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail_localize/views/edit_translation.py", line 692, in edit_translation
    location_info = get_segment_location_info(
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/wagtail_localize/views/edit_translation.py", line 318, in get_segment_location_info
    field = source_instance._meta.get_field(field_path_components[0])
  File "/home/jmunn/.virtualenvs/example-venv/lib/python3.8/site-packages/django/db/models/options.py", line 674, in get_field
    raise FieldDoesNotExist(
django.core.exceptions.FieldDoesNotExist: NewModel has no field named 'foo_description'
```

## Steps to reproduce

1. Install and setup [example repo](https://github.com/marekndt04/test_migrations) (but any basic project with `wagtail-localize` and a simple page model will do)
2. Configure a new locale
3. Create an instance of `NewModel`, saving changes as draft only (or publish first, make changes, and save as draft)
4. Create a translation of new page instance (from page explorer: more -> translate this page)
5. Publish the translation
6. Edit the definition of `NewModel`, renaming one of the fields (e.g. `description` -> `foo_description`)
7. `makemigrations` (enter `y` to indicate you renamed the field) and `migrate`
8. Attempt to visit the editor page for the translation, this should cause the error.

The error does not occur when the source page has no unpublished revisions, the warning message (`The data model for 'New model' has been changed since the last translation sync. If any new fields have been added recently, these may not be visible until the next translation sync.`) is displayed.

Syncing the translations (page explorer -> source page -> more -> Sync translated pages) stops the error from occurring, but ideally the error would be handled gracefully.

thibaudcolas:
## Description

Fixes #590 — `FieldDoesNotExist` error when editing a translation whose source page has unpublished draft changes after a migration removes/renames a field.

It felt safe to ignore those now-removed fields as part of this process. The alternative would be to show a warning asking the user to do a sync first.

<details>
<summary>AI details</summary>

The translation editor calls `get_segment_location_info()`, which does `source_instance._meta.get_field(field_path[0])` (edit_translation.py:279). When a migration renames/removes a field that an existing draft segment still references, this raises `FieldDoesNotExist`, crashing the editor with a 500.

## Changes

- Catch `FieldDoesNotExist` alongside `FieldHasNoEditPanelError` at all three call sites in `edit_translation()` (lines 674, 704, 848), so stale segments referencing removed fields are skipped instead of crashing.
- Add a regression test that creates stale `StringSegment`/`OverridableSegment` records pointing at a nonexistent field and asserts the editor returns 200 and omits those segments.

Note: this gracefully handles the stale segments (the symptom); the underlying schema-out-of-date cleanup in the data model is out of scope.

</details>

## Testing

- `testmanage.py test tests.test_edit_translation` — 88 tests pass
- `ruff check` and `ruff format` clean
- Open Code Review: 0 findings

### DIFF
--- a/src/wagtail_localize/views/edit_translation.py
+++ b/src/wagtail_localize/views/edit_translation.py
@@ -10,7 +10,7 @@ import polib
 from django.conf import settings
 from django.contrib.admin.utils import quote
 from django.contrib.auth import get_user_model
-from django.core.exceptions import PermissionDenied, ValidationError
+from django.core.exceptions import FieldDoesNotExist, PermissionDenied, ValidationError
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models, transaction
 from django.http import Http404, HttpResponse
@@ -671,7 +671,7 @@ def edit_translation(request, translation: Translation, instance):
                 segment.context.path,
                 segment.context.get_field_path(source_instance),
             )
-        except FieldHasNoEditPanelError:
+        except (FieldHasNoEditPanelError, FieldDoesNotExist):
             continue
 
         segments.append(
@@ -701,7 +701,7 @@ def edit_translation(request, translation: Translation, instance):
                 segment.context.get_field_path(source_instance),
                 widget=True,
             )
-        except FieldHasNoEditPanelError:
+        except (FieldHasNoEditPanelError, FieldDoesNotExist):
             continue
 
         segments.append(
@@ -845,7 +845,7 @@ def edit_translation(request, translation: Translation, instance):
                 segment.context.path,
                 segment.context.get_field_path(source_instance),
             )
-        except FieldHasNoEditPanelError:
+        except (FieldHasNoEditPanelError, FieldDoesNotExist):
             continue
 
         segments.append(
--- a/tests/test_edit_translation.py
+++ b/tests/test_edit_translation.py
@@ -49,6 +49,7 @@ from wagtail_localize.models import (
     OverridableSegment,
     SegmentOverride,
     String,
+    StringSegment,
     StringTranslation,
     Translation,
     TranslationContext,
@@ -500,6 +501,41 @@ class TestGetEditTranslationView(EditTranslationTestData, TestCase):
             related_object_segment["translationProgress"],
             {"totalSegments": 1, "translatedSegments": 0},
         )
+
+    def test_edit_page_translation_with_removed_field(self):
+        # Simulates a segment referencing a field that was removed/renamed by a
+        # migration after translations were created. This should be skipped
+        # instead of raising FieldDoesNotExist. See wagtail-localize#590.
+        StringSegment.objects.create(
+            source=self.page_source,
+            context=TranslationContext.objects.create(
+                object=self.page_source.object,
+                path="removed_string_field",
+                field_path="removed_string_field",
+            ),
+            order=0,
+            string=String.from_value(self.page.locale, StringValue("Old field")),
+        )
+        OverridableSegment.objects.create(
+            source=self.page_source,
+            context=TranslationContext.objects.create(
+                object=self.page_source.object,
+                path="removed_override_field",
+                field_path="removed_override_field",
+            ),
+            order=0,
+            data_json='"foo"',
+        )
+
+        response = self.client.get(
+            reverse("wagtailadmin_pages:edit", args=[self.fr_page.id])
+        )
+        self.assertEqual(response.status_code, 200)
+
+        props = json.loads(response.context["props"])
+        content_paths = [segment["contentPath"] for segment in props["segments"]]
+        self.assertNotIn("removed_string_field", content_paths)
+        self.assertNotIn("removed_override_field", content_paths)
 
     def test_page_chooser_widgets(self):
         home_page_with_specific_type = self.home_page.add_child(


rypptc (commented):
I reproduced this on the demo site and then walked the flow with a debugger. Here's what I found.

- The error shows up after renaming a field even with both pages (en/fr) published and `has_unpublished_changes=False`, which suggests the issue isn't limited to draft sources. What it needs is a segment pointing at a field that no longer resolves. That's what I see on the demo today — the behaviour back in 2022, when the issue was filed, may have been different.

- The PR catches `FieldDoesNotExist` in the `except` of the three loops that build the segment list (`edit_translation.py:674`, `:704`, `:848`); the `continue` skips that segment and the exception doesn't propagate. On the demo the editor goes from 500 to 200: it shows the nine valid string segments and leaves out the invalid one. With the request no longer interrupted, execution reaches `edit_translation.py:877`, where `schema_out_of_date()` returns `True` and the warning about the data model having changed since the last sync is shown.

- With the patch applied, but before syncing, publishing and preview still fail with `FieldDoesNotExist` in `ingest_segments()`. Neither goes through those three loops: they go through `segments/ingest.py:293`, which does its own field lookup, unguarded. I tried it by hitting Publish in the admin: 500, same exception. Preview fails the same way. Translations saved segment by segment aren't lost — I checked by saving one and then publishing, and the 500 doesn't delete the row — but the page can't be published or previewed. The warning the editor now shows suggests syncing.

- The issue suggests that syncing solves this, and it does unblock publishing, but it doesn't reassociate the existing translation. `update_from_db()` re-reads the source page and updates the stored copy; then `refresh_segments()` re-extracts the segments from that copy: it creates a `TranslationContext` with the new name and its segment, and deletes the old segment (`models.py:579`, which deletes segments, never contexts). I measured it on the demo, before and after, inside a rolled-back transaction.

  For the subtitle segment:

  ```
  before sync
    TranslationContext:      field_path='subtitle'
    StringTranslation (fr):  "L'art de la boulangerie"

  after sync
    TranslationContext:      field_path='foo_subtitle'
    StringTranslation (fr):  (none)
  ```

  What gets deleted is the segment, which is the link. The old `TranslationContext` and `StringTranslation` remain in the database, but the source no longer has a segment referencing that context, so the editor doesn't pick up the previous translation. The object's contexts go from 15 to 16 — the old one isn't deleted, a new one is added.

  What this means for a translator: after syncing, the field shows up untranslated even though the English source text is exactly the same. Syncing doesn't carry the translation over, so it has to be entered again.

- The PO file that gets downloaded includes the stale context (`msgctxt "subtitle"`), so an external translator can end up translating an entry that no longer has anywhere to go.

The change looks right to me and it improves things: today the editor doesn't open and with the patch it does.

Publish and preview go through a different code path, so I'd see that as a separate issue rather than something to solve here. I'm happy to open it with the reproduction and the traceback.

And on the underlying problem, to consider separately: the equivalence between the old and the new name is known to the developer writing the migration, and the editor can no longer infer it when it runs into the old name months later. That equivalence is recorded in the `RenameField` operation, so a data migration alongside it could adapt the translation data.

A local test suggests one way this could work: for this simple rename, updating `path` and `field_path` and recomputing `path_id`, while preserving the context's primary key, keeps the segment and its translation attached to it. With that, the editor opens, publishing writes the translation into the renamed field (I checked the stored value, not just that it didn't raise), preview shows it, and a later sync reuses that context instead of creating a new one. It could be a starting point for a Localize-specific migration operation.

On how it could reach the developer: `post_migrate` receives the migration `plan`, so a receiver could spot a `RenameField` on a translatable model, check whether any segments still point at the old name, and say so. The repair itself would probably be better as an explicit operation the developer adds next to the `RenameField` — it shows up in the diff, it can define its own reversal, and it doesn't rewrite data during a deploy on its own.

That was a single test on one simple rename, so there'd be edge cases to work out — collisions with existing contexts, structured fields like StreamField — before it's more than an idea. Would it be worth looking into?

PR #958, opened 2026-08-25
Use the current mkdocstrings options key in the models reference pages

### LINKED ISSUE #957: Reference pages under `docs/ref/models/` ignore their member lists
### Issue summary

The five pages under `docs/ref/models/` all point at `wagtail_localize.models`, and each declares which classes to publish. None of it applies: every page renders the whole module, so the five are nearly identical. See [ref/models/translation-management](https://wagtail-localize.org/stable/ref/models/translation-management/), which declares four classes and renders classes belonging to the other four pages.

The declarations sit under `selection:`, which mkdocstrings deprecated in 0.19 and removed in 0.23 in favour of `options:`. `mkdocs.yml` was migrated; the page-level directives were not. That was harmless while the project stayed on an older version, and only took effect once the docs dependencies moved past 0.23. An unrecognised key is ignored silently, so nothing warned.

### Proposed solution

Rename the key in those five files, leaving `members` and `filters` untouched.

### Steps to reproduce

1. Open [`docs/ref/models/translation-management.md`](https://github.com/wagtail/wagtail-localize/blob/main/docs/ref/models/translation-management.md). It declares four classes: `TranslatableObject`, `TranslationSource`, `Translation`, `TranslationLog`.
2. Open the page it generates: https://wagtail-localize.org/stable/ref/models/translation-management/
3. The page renders 24 members — 19 classes plus a few functions, methods and an attribute. Among them `StringSegment`, `LocaleSynchronization` and `StringTranslation`, which belong to the other four pages, and internal ones no page declares such as `POImportWarning`.

The same applies to the other four pages under `docs/ref/models/`.

This was also reproduced and fixed locally, on the versions currently pinned (mkdocstrings 0.28.3, mkdocstrings-python 1.16.12):

    python3 -m venv .venv-docs
    .venv-docs/bin/pip install -e ".[documentation]"
    .venv-docs/bin/mkdocs serve

The local build rendered the same 24 members as the published page, so the cause is the repository's configuration rather than anything about the publishing host. Renaming `selection:` to `options:` in that one file, with the server running, made the page reload showing exactly its four declared classes — nothing else was changed.

### Can be reproduced

Yes, on the published docs site and on a local build.

### Working on this

A PR follows, limited to renaming the key. The member lists stay as they are.

rypptc:
Fixes #957

### Description

The five pages under `docs/ref/models/` declare their members under `selection:`, which mkdocstrings removed in 0.23. Renaming it to `options:` is the whole change — the `members` and `filters` lists are untouched.

Checked with a local build: each page now renders exactly what it declares — 4, 4, 5, 2 and 1 members respectively, against 24 on every page before.

### AI usage

Used AI assistance to investigate the cause, apply the rename across the files, and draft this description. I reproduced the bug and verified the result with a local build myself.

### DIFF
--- a/docs/ref/models/overridable-segments.md
+++ b/docs/ref/models/overridable-segments.md
@@ -19,7 +19,7 @@ style E stroke-dasharray: 5 5
 ```
 
 ::: wagtail_localize.models
-    selection:
+    options:
         members:
             - SegmentOverride
             - OverridableSegment
--- a/docs/ref/models/settings.md
+++ b/docs/ref/models/settings.md
@@ -12,7 +12,7 @@ style B stroke-dasharray: 5 5
 ```
 
 ::: wagtail_localize.models
-    selection:
+    options:
         members:
             - LocaleSynchronization
         filters:
--- a/docs/ref/models/translatable-segments.md
+++ b/docs/ref/models/translatable-segments.md
@@ -22,7 +22,7 @@ style I stroke-dasharray: 5 5
 ```
 
 ::: wagtail_localize.models
-    selection:
+    options:
         members:
             - SegmentOverride
             - StringSegment
--- a/docs/ref/models/translation-management.md
+++ b/docs/ref/models/translation-management.md
@@ -18,7 +18,7 @@ style F stroke-dasharray: 5 5
 ```
 
 ::: wagtail_localize.models
-    selection:
+    options:
         members:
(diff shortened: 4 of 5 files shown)


thibaudcolas (approved):
thank you! I’m surprised this wouldn’t have failed the build altogether.

PR #959, opened 2026-09-01
Use wagtail-localize's features in the demo site

rypptc:
Follows #942, which added the demo site. This makes its models translatable, ships one translated post in the initial data, and adds tests and a README, following the suggestions in #928.

### Description

Adds `TranslatableMixin` to `Person` and to `BlogPersonRelationship`, so a translated post links to the translated author. Without it on the relationship, the link row is copied unchanged and the French post lists the English author.

First and last names are synchronised and the job title is translatable, since a name is the same in both languages and a job title is not.

Only one post is translated, so `/fr/` and `/fr/blog/` remain aliases of the English pages.

The DeepL translator is now only configured when `DEEPL_AUTH_KEY` is set. The translation editor shows the "Translate with DeepL" action whenever a translator is configured, without checking the key, and without a key pressing it raises `AttributeError` before any request reaches DeepL.

The language switcher is not part of this, pending https://github.com/wagtail/bakerydemo/pull/760.

### AI usage

Used AI assistance throughout this change, including the French translations, and to draft this description. I reviewed, tested, and understood the result.


.github/workflows/test.yml
```diff
@@ -141,7 +141,7 @@ jobs:
       - name: Check for missing demo migrations
         run: python demo/manage.py makemigrations --check --noinput
       - name: Run the demo tests
-        run: python demo/manage.py test home
+        run: python demo/manage.py test home blog
```
thibaudcolas, line 144:
```suggestion
        run: python demo/manage.py test
```

demo/blog/models.py
```diff
@@ -51,6 +54,12 @@ class Person(
         related_name="+",
     )
 
+    # Names are the same in every language; the job title is not.
+    override_translatable_fields = [
+        SynchronizedField("first_name"),
+        SynchronizedField("last_name"),
+    ]
```
thibaudcolas, line 61: very nice! see too few projects out there doing this properly

demo/blog/management/commands/load_initial_data.py
```diff
@@ -824,3 +937,19 @@ def load_image(spec: dict[str, Any]) -> Image:
         self.stdout.write(
             self.style.SUCCESS("Homepage updated with hero and blog section.")
         )
+
+        french, _created = Locale.objects.get_or_create(language_code="fr")
+        post = BlogPage.objects.filter(
+            slug=TRANSLATED_POST_SLUG, locale=Locale.get_default()
+        ).first()
+        if post:
```
thibaudcolas, line 945: redundant conditional?

thibaudcolas (approved):
great!

demo/blog/management/commands/load_initial_data.py
rypptc: There is an edge case. The user runs load_initial_data once, then deletes the post, and later runs load_initial_data again. In this case, the post would not exist.

rypptc:
There was an error in the migration. I found it when updating a demo that already has content. I updated the migration file using BootstrapTranslatableModel to fill the locale and the translation key of each existing row, before the unique constraint is applied. I also added a test that creates the rows with the old models and migrates them forward.

